PR #21, opened 2020-11-03
update release process

uturunku1:
# Issue
Previously our release process was easily triggered with the push of a git tag. However, last-minute yesterday, I was concerned that anybody with access to our repo could accidentally push a tag. Which it is possible, as I have read that contributors do request GitHub to have a feature like bitbucket and gitlab to make the usage of tags more restricted. Nonetheless, if an internal person to New Relic were to push a tag accidentally, I think they would warn us immediately and the process is easily reversible. That explains why some of our other repos under this org are ok using the simple push of tag to trigger their release process.
Still, yesterday I spend most of the day attempting to find a way to make the release process just a little more restrictive, or better said add more effort to it, so it makes it hard to trigger a release by accident. I settled for the solution that is implemented in my changes.

# Goals
Find a different trigger for the release process that is more manual and less easy to push.

# Implementation Details
I have added new instructions to Release-Runbook.md that should explain the implementation steps behind the new release workflow.

One important note is that you'll notice I made a small change in lint.yml. This is completely unrelated to this PR. However, I'd still like to merge this change: It will turn off the lint when merging a PR into the main branch(which we have seen it looks into the diff and alerts on old lint issues). I am letting golangci-lint only run when someone has a PR opened so it will only look at the current changes coming in, and alert on NEW lint issues.

# How to Test
We'll need to approve this PR so I can test the new workflow.


GabrielSamp (approved):
Looks good!

PR #34, opened 2020-11-06
Refactor AppName task

uturunku1:
# Issue
Base/Env/AppName task was yelling too much at the customer. It is unneeded to give a warning status when a customer has at least provided an app name for the main app name setting. The other app name settings that show up in the same config file are optional. 
Example of how a NR config file has multiple settings for app name but only the first one is required:

```
/common/app_name: Luces-sqs-java
 /development/app_name: My Application (Development)
 /production/app_name: Luces-sqs-java
 /staging/app_name: My Application (Staging)
 /test/app_name: My Application (Test)
```

Additionally, this task is not looking at the env var setting for the app name, which is super relevant because if it is set, then it will overwrite all config files settings (except for Python but we'll have to ignore that case for now. Reach out to me for a background story on this), and it is wrong of us to give this failure status the customer that they are missing an app name in their config file when they clearly have set it in the env var:

```
No New Relic app names were found. Please ensure an app name is set in your New Relic agent configuration.
```

# Goals
1. Check for env var and if its set, then ignore the app name settings in the config file
2. Look at the main app name setting in the config file and ignore the optional settings for other app names, such as production, test, and staging environment.

# How to Test
I adjusted the unit tests after the changes, and I added a new one for checking for env vars.


tasks/base/config/appname.go
```diff
@@ -107,43 +129,75 @@ func (t BaseConfigAppName) Execute(options tasks.Options, upstream map[string]ta
...
+		/*
+			Only grab the first appname key found as this is the main required for an app to start reporting. The other appname keys are optional. Example:
+				/common/app_name: Luces-sqs-java
+				/development/app_name: My Application (Development)
+				/production/app_name: My Application (Production)
+				/staging/app_name: My Application (Staging)
+				/test/app_name: My Application (Test)
+		*/
+		if len(foundKeys) > 0 {
+			key := foundKeys[0]
```
MichelLosier, line 170: Is the first index of foundKeys deterministic to always being the primary app_name configuration? Is there a way to target this specifically?

tasks/base/config/appname.go
```diff
@@ -71,32 +97,28 @@ func (t BaseConfigAppName) Execute(options tasks.Options, upstream map[string]ta
...
 		}
 	}
-	
-	defaultNameMatches := ""
 
-	appNameInfoFromConfig := getAppNamesFromConfig(configElements, appNameConfigKeys)
-	if len(appNameInfoFromConfig) == 0 {
+	appNameInfosFromConfig := getAppNamesFromConfig(configElements)
+
+	if len(appNameInfosFromConfig) == 0 {
```
MichelLosier, line 103: Do we want to validate if we find a config file without an app name? This is outside of the scope of this PR and part of the original behavior of the task, so this is not a blocker, but just wanted to raise the question. Might be better suited for agent specific config validation.

tasks/base/config/appname.go
```diff
@@ -107,43 +129,75 @@ func (t BaseConfigAppName) Execute(options tasks.Options, upstream map[string]ta
...
+				/staging/app_name: My Application (Staging)
+				/test/app_name: My Application (Test)
+		*/
+		if len(foundKeys) > 0 {
+			key := foundKeys[0]
+			if !key.IsLeaf() {
+				for _, child := range key.Children {
+					appName := child.Value()
+					result = append(result, AppNameInfo{
+						Name:     appName, // should we sanitize this?
```
MichelLosier, line 175: Can probably drop this comment here and on 184. Seems to be an artifact from orig. implementation.

tasks/base/config/appname.go
```diff
@@ -50,13 +65,24 @@ func (t BaseConfigAppName) Explain() string {
...
 // Execute - The core work within each task
 func (t BaseConfigAppName) Execute(options tasks.Options, upstream map[string]tasks.Result) tasks.Result {
 
-	// check to see if upstream was successful, exit if not
+	appNameInfoFromEnvVar := getAppNameFromEnvVar(upstream)
+	//We can have an early exit because this env var will overwrite all config files setting for app name, except for Python
+	if len(appNameInfoFromEnvVar.Name) > 0 {
+		return tasks.Result{
+			Status:  tasks.Success,
+			Summary: fmt.Sprintf("A unique application name was found through the New Relic' App name environment variable: %s", appNameInfoFromEnvVar.Name),
```
MichelLosier, line 80: Can drop the apostrophe after `New Relic'`

MichelLosier (changes requested):
Really like the experience change proposed here. Looks good!

tasks/base/config/appname.go
uturunku1: Agree. Though it does seem so far that, at least for Java and Node, that the order can be predicted, I can attempt to get the primary app_name by the path that differentiates it from the other non-primary names: `/common/` (fullpath: /common/app_name)

tasks/base/config/appname.go
uturunku1: That is something to think about, and I'll have to research among TSEs to see if there is much difference in the way we validate app names between agents. So far to me, it looks like this validation is in the right place, and that we can validate app name in the `Base` level because it is applicable to all agents that they have at least a primary app name, and if they don't we warn the customer about it.

MichelLosier (approved):
Good to go! 🏇  Thanks!

Issue #32, opened 2020-11-05
Base/Env/InitSystem doesn't detect OpenRC for Alpine Linux / Busybox


uturunku1:
## Summary

```
Base/Env/InitSystem
Summary
Unable to parse init system from: /bin/busybox
```

https://wiki.gentoo.org/wiki/OpenRC#busybox

In the code: https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/base/env/initSystem.go#L69

uturunku1:
The issue was addressed here: https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/base/env/initSystem.go#L91

PR #35, opened 2020-11-06
detect OpenRC init system for busybox

uturunku1:
# Issue

It's addressing this GitHub issue: https://github.com/newrelic/newrelic-diagnostics-cli/issues/32

To summarize, our customers occasionally will get this error in the nrdiag output on their terminal:

```
Base/Env/InitSystem
Summary
Unable to parse init system from: /bin/busybox
```

A few customers and TSEs, have reached out to understand if this error is something they should concern about and fix on their end. Meaning that the error status, instead of filling the purpose of letting them know that nrdiag had an issue finding the initsystem, makes them think that they (the customer) caused the error and they are at fault.

# Goals

1. Changing Status: `tasks.Error` to `tasks.None` if we could not determine the type of Linux system.
2. Adding a check for busybox (a popular type of linux flavor, that uses the init system and is compatible with it)

# Implementation Details

# How to Test

### DIFF
--- a/tasks/base/env/initSystem.go
+++ b/tasks/base/env/initSystem.go
@@ -3,6 +3,7 @@ package env
 import (
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/newrelic/newrelic-diagnostics-cli/tasks"
 )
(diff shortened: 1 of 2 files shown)


punkael (approved):
Assuming busybox is used in the path, this looks great.

PR #36, opened 2020-11-08
Collect sys props and use them in relevant tasks

uturunku1:
# Issue

We have tasks that look into config file settings and environment variables to validate if a customer's configuration has been appropriately implemented, for example, Base/Config/AppName, Base/Config/ValidateLicenseKey, and Base/Config/ValidateHSM. The issue is that these tasks would possibly give a false positive result if the configuration has applied through JVM System Properties (in the case of the Java agent) because we do not collect and look at system properties yet, only env vars and settings in config files. 

# Goals

1.Add a task for collecting system properties: Base/Env/CollectSysProps

2.Use the upstream payload from Base/Env/CollectSysProps in tasks that validate Java Agent configuration settings. Those tasks are:

Base/Config/Collect
Base/Config/AppName
Base/Config/LicenseKey
Base/Log/Loghelpers
Base/Log/Copy

I was going to include our high-security validation task (Base/Config/ValidateHSM) in this list but while checking in docs, tickets, slack, and consulting with a TSE, it doesn't seem that `-Dnewrelic.config.high_security` is used among customers. Is either deprecated or is not used much in comparison to the env var or config file options for high-security mode. That's why I deprioritize implementing system property into this task.

# Implementation Details

Before you go into reviewing the "Files changed" section/tab above(which does not display the changes in chronological order), I recommend starting to review the implementations in the following order:
1.CollectSysProps.go : 
https://github.com/newrelic/newrelic-diagnostics-cli/blob/b4b1e1bec24b6aac4935adb54a2298161705e038/tasks/base/env/collectSysProps.go

2.TasksHelpers.go: 
https://github.com/newrelic/newrelic-diagnostics-cli/blob/b4b1e1bec24b6aac4935adb54a2298161705e038/tasks/taskHelpers.go#L772

3.Any of the tasks that were modified within this path: tasks/base/config/
The order does not matter. You can skip reviewing the _test.go files if you wish, but it may be worth just to take a quick peek, so you get an idea of what I expect the behavior of the task to be after my changes

# How to Test

Most of the tasks where I have implemented changes to validate for system properties, already have a bunch of units test that should confirm that expected behavior after those changes. Please check those unit tests as they are very thorough and document really well what I expect those changes to accomplish.
On the other hand, there are a couple of tasks that do not have a file for unit tests yet (and I am not adding that file at least for this PR or MMF story), so if you can smoke test those couple of tasks (Base/Config/Collect and Base/Log/Copy) by running them in a real Java app environment that sets `-Dnewrelic.config.file` and `-Dnewrelic.log`, that'd be a great test! I have smoke test them myself in darwin but is always nice to try it out in different environments (Linux or windows). 

This is the way you can smoke test in your environment:
-First, clone this repo and then check-in into this specific branch (`git checkout collectSysProps`)
-Then you can run `go build`, or you can run `./scripts/build.sh`, which will build the binary for all operating systems.
-`go build` will build the binary `newrelic-diagnostics-cli` in the current work directory. But `./scripts/build.sh` will build the binaries inside of  a bin/ directory.
-Now you can bring your binary and drop it into your java app directory. You can either run all java tasks with `./newrelic-diagnostics-cli -y -suites java` 
Or you can target the specific task by running `./newrelic-diagnostics-cli -t Base/Config/Collect`

Note: the `-y` option is for saying yes to all prompts nrdiag will ask before collecting a file (security reasons)

### DIFF
--- a/tasks/base/config/appname.go
+++ b/tasks/base/config/appname.go
@@ -65,8 +65,8 @@ func (t BaseConfigAppName) Explain() string {
 func (t BaseConfigAppName) Dependencies() []string {
 	return []string{
 		"Base/Config/Validate",
-		"Base/Env/collectEnvVars",
-		"Base/Env/collectSysProps",
+		"Base/Env/CollectEnvVars",
+		"Base/Env/CollectSysProps",
 	}
 }
 
@@ -78,7 +78,7 @@ func (t BaseConfigAppName) Execute(options tasks.Options, upstream map[string]ta
 	if len(appNameInfoFromEnvVar.Name) > 0 {
 		return tasks.Result{
 			Status:  tasks.Success,
-			Summary: fmt.Sprintf("A unique application name was found through the New Relic' App name environment variable: %s", appNameInfoFromEnvVar.Name),
+			Summary: fmt.Sprintf("An application name was found through the New Relic' App name environment variable: %s", appNameInfoFromEnvVar.Name),
 			Payload: []AppNameInfo{appNameInfoFromEnvVar}, //though is a single item, we still add them to a slice of AppNameInfo to stay consistant with a future upstream payload type assertion
 		}
 	}
@@ -89,7 +89,7 @@ func (t BaseConfigAppName) Execute(options tasks.Options, upstream map[string]ta
 	if appname != "" {
 		return tasks.Result{
 			Status:  tasks.Success,
-			Summary: fmt.Sprintf("A unique application name was found through the a New Relic system property: %s", appname),
+			Summary: fmt.Sprintf("An application name was found through the a New Relic system property: %s", appname),
 			Payload: []AppNameInfo{AppNameInfo{Name: appname, FilePath: appNameSysProp}},
 		}
 	}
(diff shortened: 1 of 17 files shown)


tasks/base/config/appname.go
```diff
@@ -131,12 +143,13 @@ func (t BaseConfigAppName) Execute(options tasks.Options, upstream map[string]ta
 
 	return tasks.Result{
 		Status:  tasks.Success,
-		Summary: fmt.Sprintf("%s unique application name(s) found.", strconv.Itoa(len(appNameInfosFromConfig))),
+		Summary: fmt.Sprintf("%s unique application name(s) found: %s", strconv.Itoa(len(appNameInfosFromConfig)), appNameInfosFromConfig[0].Name),
```
MellowMarble, line 146: Just optimization but, since len() returns an int ([reference](https://golang.org/pkg/builtin/#:~:text=The%20len%20built%2Din%20function,len(v)%20is%20zero.)) we should be able to bypass the string conversion and just use something like %d for the first value

uturunku1: Good catch! I will apply the suggested change.

MellowMarble:
Hi @uturunku1 - I ran a test locally and just found a couple of issues:

1. The Diagnostics CLI does correctly detect the JVM Argument setting a custom log location `-Dnewrelic.config.logfile` however it still ends up gathering the actual log file from the default location of "/newrelic/logs/newrelic_agent.log"
2. It looks like the logic for capturing JVM arguments is space-delimited and causes and issue when detecting a configured application name (via JVM args) that includes a space, such as: `-Dnewrelic.config.app_name="My Application"` - In that case it will provide a result showing `-Dnewrelic.config.app_name="My"`

MellowMarble (approved):

PR #37, opened 2020-11-11
Updated tool naming to be consistent

GabrielSamp:
# Issue
The Diagnostics CLI was referred to by a number of different names in documents and summaries 
# Goals
Consistent naming of the tool
# Implementation Details
changed name directly in docs and a few places outside tasks. In tasks created a const named ThisProgramFullName in tasks.go to use in summaries and other task output as needed
# How to Test
Should just be standard tests

### DIFF
--- a/docs/Release-Runbook.md
+++ b/docs/Release-Runbook.md
@@ -1,4 +1,4 @@
-# Diagnostics CLIrelease runbook - Internal usage only
+# Diagnostics CLI release runbook - Internal usage only
 
 This assumes the main branch is in a state ready for release. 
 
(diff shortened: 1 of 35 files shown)


docs/Release-Runbook.md
```diff
@@ -1,4 +1,4 @@
-# NR Diag release runbook - Internal usage only
+# Diagnostics CLIrelease runbook - Internal usage only
```
uturunku1, line 1: There is a spacing missing here.

uturunku1 (changes requested):
Sweet! Thanks for the quick changes. I'm requesting one single change before we can merge this🐎

docs/Release-Runbook.md
GabrielSamp: added space

uturunku1 (approved):

PR #38, opened 2020-11-13
Copy over previous work and browser fixtures

CherylFrankenfield:
# Issue

BrowserAgentGetSource task collects the Browser agent script found in the HTML source page. However, the task should not collect the Browser script if this is found outside of the Head tag element because it hides a potential issue on the customer's end.

# Goals
Improve regex to detect when the Browser script is not in the Head tag.
If it is found somewhere else that it is not the Head, then we should notify the customer about instrumentation issues.

# Implementation Details
The changes have been made in Tasks/Browser/Agent/GetSource.go
We have added some unit tests
Finally, we have enabled the browser suite that previously was commented out in suitesDefinitions.go
 
# How to Test
To test this task, you can build the go binary  with `go build` against this branch, and then run `./newrelic-diagnostics-cli -browser-url https://someurlthathasthenewrelicbrowserscript -suites browser`
In this ticket, you can find a URL that "technically" has injected the browser script, but has done it following bad practices: https://newrelic.zendesk.com/agent/tickets/443022 . So we would expect Diagnostics CLI to give a `failure` status.
This ticket(https://newrelic.zendesk.com/agent/tickets/423509) can provide a URL that has placed the browser script in the right place, so the task should provide a `success` status.


docs/Dependency-Injection.md
```diff
@@ -100,7 +100,7 @@ Here we have a struct field name `httpGetter` with the type: `httpRequestFunc`.
 
 ### (2) Attach our `makeCollectorRequest()` to the task struct as a struct method (so it can reference its dependencies defined as struct fields)
 
-Next, we attach our `makeCollectorRequest` function to our task `struct` so it can reference dependencies that are fields on the struct instance. We do this by adding the struct as a receiver (`(p BaseCollectorConnect)`) in the function declaration:
+Next, we attach our `makeCollectorRequest` function to our task `struct` so it can reference dependencies that are fields on the struct instance. We do this by adding the struct as a receiver (`(p BaseCollectorConnectUS)`) in the function declaration:
```
uturunku1, line 103: good catch. We do not have a task name BaseCollectorConnect

tasks/browser/agent/agent.go
```diff
@@ -9,6 +9,6 @@ import (
 func RegisterWith(registrationFunc func(tasks.Task, bool)) {
 	log.Debug("Registering Browser/Agent/*")
 
-	registrationFunc(BrowserAgentDetect{}, false)
-	registrationFunc(BrowserAgentGetSource{}, false)
+	registrationFunc(BrowserAgentDetect{}, true)
```
uturunku1, line 12: I can't believe this is the first time we enable these task by default but it was about time

tasks/browser/agent/detect.go
```diff
@@ -29,10 +29,10 @@ func (t BrowserAgentDetect) Identifier() tasks.Identifier {
 // Explain - Returns the help text for each individual task
 func (t BrowserAgentDetect) Explain() string {
 	return "Detect New Relic Browser agent from provided URL"
-	// gorun -t browser/agent/detect -v -o Browser/Agent/Detect.url=http://localhost:3000
+	// ./nrdiag -browser-url http://thecustomers-website-url --suites browser
```
uturunku1, line 32: I have smoke tested this command with different URLs and it works great!

uturunku1 (approved):
The big change that this PR brings is that we are going to use unit tests and fixtures instead of the previous integration tests we use to have around Browser.
Also, we are enabling the tasks to run by default as long as the user passes into the command line a URL pointing to the site that has the Browser script

PR #41, opened 2020-11-24
Validate nrjmx mbeans

uturunku1:
# Issue
Not an issue, we are adding new features that I'll describe in the next section. 
# Goals
1)Adding a task Infra/Env/NrjmxBeans: Validate that the mbeans that the customer wants to query with the JMX integration are available for the integration to pull from.
2)Add two new fields to the struct JmxConfig found in the task Infra/Config/ValidateJMX: This task is mainly used by TSEs to troubleshoot that a JMX server cannot connect to our JMX integration, and not so much for a customer to understand WHY a connection is failing. This is because we only output the error found, but we do not judge what the error means.
With that same purpose of using this task only to provide data to TSEs to troubleshoot, I have added two new fields to this task struct that contains information about the Java version and the cmdline arguments for JMX. This specific information was requested by infra TSEs after I met with them and they said they use those 2 pieces of information to troubleshoot alongside the error that nrdiag provides. This information will only be provided to TSEs and not customers as it only shows up in the payload result, and not in the summary result.

# Implementation Details

1)Infra/Env/NrjmxMbeans:
-It will parse through the JMX integration metrics files listed under in collection_files field found in the jmx-config.yml
-It will grab the all queries defined in those metrics files and run them against the command `{queryname} | nrjmx -H %s -P %s -v -d -`
-If any of the queries return an empty object, it will warn the customer about this query not being available to the integration

2)Infra/Config/ValidateJMX:
-To get data the java version, I am leveraging the Java/Env/Version task to get the payload string
-To get the cmdline arguments for JMX, I am running this command `ps -ef | grep jmx` and converting the output []bytes to a string. Then adding that string to our result payload for TSE to interpret that string to the best of their knowledge. 

# How to Test
I have added a couple of unit test to the new task: Infra/Env/NrjmxMbeans.
I have also smoke tested in a linux environment that has JVM processes running and one of them had JMX enabled. I installed and configured the JMX integration to point to the incorrect port for the JMX server just to see the task fail with a very common mistake customers make. 

If you are interested in smoke testing this task, you'll have to follow these instructions to set the JMX integration: https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/jmx-monitoring-integration

To enable JMX metrics on your testing app, you can pass arguments like these when you start up your java app: 
java-jar -Dcom.sun.management.jmxremote.password.file=jmxremote.password-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.authenticate=true-Dcom.sun.management.jmxremote.port=9010-Dcom.sun.management.jmxremote.ssl=false -pathtoyourapp.jar

Then you'll have to checkout into this branch and run ./scripts/build.sh to get the linux binary and drop the binary in your JMX server. This task does not run yet on windows. Though I am planning to on my next PR to fix a bug in ValidateJMX task that does not allow for this task to run on Windows.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,5 @@ require (
 	github.com/stretchr/testify v1.6.1 // indirect
 	golang.org/x/sys v0.0.0-20201013132646-2da7054afaeb
 	gopkg.in/cheggaaa/pb.v1 v1.0.28
-	gopkg.in/yaml.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 )
--- a/go.mod
+++ b/go.mod
@@ -18,5 +18,6 @@ require (
 	github.com/stretchr/testify v1.6.1 // indirect
 	golang.org/x/sys v0.0.0-20201013132646-2da7054afaeb
 	gopkg.in/cheggaaa/pb.v1 v1.0.28
+	gopkg.in/yaml.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 )
--- a/tasks/infra/config/validateJMX.go
+++ b/tasks/infra/config/validateJMX.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"os/exec"
 
 	log "github.com/newrelic/newrelic-diagnostics-cli/logger"
 	"github.com/newrelic/newrelic-diagnostics-cli/tasks"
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -18,5 +18,6 @@ require (
 	github.com/stretchr/testify v1.6.1 // indirect
 	golang.org/x/sys v0.0.0-20201013132646-2da7054afaeb
 	gopkg.in/cheggaaa/pb.v1 v1.0.28
+	gopkg.in/yaml.v2 v2.3.0
```
MichelLosier, line 21: I see we add back in yaml.v2 as a dependency next to yaml.v3. Is there a reason we can't use yaml.v3 in these tasks?

tasks/infra/config/validateJMX.go
```diff
@@ -16,22 +15,17 @@ const defaultJMXport = "9999"
...
-	User            string
-	Password        string
-	CollectionFiles string
+	Host                  string
+	Port                  string
+	User                  string
+	Password              string
+	CollectionFiles       string
+	JavaVersion           string
+	JmxProcessCmdlineArgs string
```
MichelLosier, line 28: Is it possible for to collect sensitive information in the cmd line args of the jmx process - like auth / keystore password? If so, we may want to flag this task with a y/n prompt, or scrub values from known sensitive args.

MichelLosier (changes requested):
This is a great add, thanks for describing the scope of this in detail along with the background on what its solving here!

Have just a few comments to look at.

In addition, I wonder if its worth considering using the process package from:
https://github.com/shirou/gopsutil
which we already have as a dependency on this project. This might help with performing the same kind of process lookup in different environments.

go.mod
uturunku1: sneaky little dependency! I had removed it from `go.mod` and ran `go mod tidy` file but then, of course, Go modules brought it back because I had forgotten to remove it from the import section on top of the file task😆

tasks/infra/config/validateJMX.go
uturunku1: Yup, that's a good call. I'm removing the argument values that possibly point to sensitive data. Though, after reading the oracle docs, it seems like at most they can pass references to files where they hide that sensitive data, such as `-Dcom.sun.management.jmxremote.password.file=/pathtofilewhereihidemyauthenticationinfo`. But you never know, so I'm still redacting those values.

MichelLosier (approved):

Issue #1, opened 2020-10-18
InfraConfigValidateJMX task does not run in windows


uturunku1:
## Description

When Infra/Config/ValidateJMX task runs on a windows environment, it fails at this line: 
`echo "*:type=*,name=*" | nrjmx -hostname 127.0.0.1 -port 9999 --verbose true`
https://github.com/newrelic/NrDiag/blob/main/tasks/infra/config/validateJMX.go#L184

## Steps to Reproduce
1. Remove this line to get this task to run on windows: https://github.com/newrelic/NrDiag/blob/main/tasks/infra/config/config.go#L24
2. Then run as admin the cmd  `./nrdiag -t infra/config/ValidateJMX` to only see this task in action and filter out others tasks
3. Then you'll notice the following error:

```
Error connecting to local JMXServer:
exec: "echo": executable file not found in %PATH%
```

uturunku1:
https://github.com/newrelic/newrelic-diagnostics-cli/pull/42

PR #42, opened 2020-11-30
Issue 1/update infra config validate jmx forwindows

uturunku1:
# Issue
This PR addresses the issue listed and described here: https://github.com/newrelic/newrelic-diagnostics-cli/issues/1

The first part of this cmd (right before the pipe) fails in windows `echo "*:type=*,name=*" | nrjmx -hostname 127.0.0.1 -port 9999 --verbose true` with the error `"echo": executable file not found`
This makes sense because we are using the go built-in function exec.Command (https://golang.org/pkg/os/exec/#Command), which expects to take the name of an executable as the first parameter followed by a list of arguments that get passed to that executable `func Command(name string, arg ...string)`. 

So our cmd above runs like this: `exec.Command("echo", ":type=*,name=*")`
However, "echo" is not an executable but built-in command in Powershell.

# Goals
1. Get Infra/Config/ValidateJMX to run in Windows without throwing an error for "echo: executable not found"
2. Add Infra/Env/NrjmxMbeans task to the list of tasks that run in Windows

# Implementation Details

1. To bypass the error in ValidateJMX, I am passing `cmd.exe` (which is the windows shell executable) as the first argument passed to exec.Command, and then, I am passing the built-in commands and parameters:
`exec.Command("cmd.exec", "/C", "echo", "*:type=*,name=*")`
Based on windows documentation:  `/C`  Carries out the command specified by string and then terminates

2. Because NrjmxMbeans task was also relying on running "echo" as the first argument passed in `exec.Command`, I had to make similar adjustments as the one above.

# How to Test
To tests these tasks, you should checkout into this branch and build the windows binary `GOOS=Windows go build`. Then you can run it on a windows vm that has installed the JMX integration for windows: https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/jmx-monitoring-integration#windows-install
No need to enable jmx metrics in a java app to see this task in action because as soon as nrdiag detects the new relic configuration file for jmx, infra/Config/ValidateJMX will run, and you will not see "echo" cmd fail anymore. At most you'll see the status.Failure with the summary: cannot connect to JMX server. And that is because you have not configured your jmx server to listen to the same port that is listed in the jmx config file.
Now, if you do want to be able to run infra/Env/NrjmxMbeans, in that case, you will have to enable jmx metrics because this task only runs after ValidateJMX task has succeeded at connecting to a jmx server. This is an example of how I enabled jmx metrics for my java app:

```
java-jar -Dcom.sun.management.jmxremote.password.file=jmxremote.password-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.authenticate=true-Dcom.sun.management.jmxremote.port=9010-Dcom.sun.management.jmxremote.ssl=false -pathtoyourapp.jar
```

### DIFF
--- a/tasks/infra/config/config.go
+++ b/tasks/infra/config/config.go
@@ -20,10 +20,7 @@ func RegisterWith(registrationFunc func(tasks.Task, bool)) {
 		runtimeOS: runtime.GOOS,
 	}, true)
 	registrationFunc(InfraConfigIntegrationsValidateJson{}, true)
-
-	if runtime.GOOS != "windows" {
-		registrationFunc(InfraConfigValidateJMX{
-			mCmdExecutor: tasks.MultiCmdExecutor,
-		}, true)
-	}
+	registrationFunc(InfraConfigValidateJMX{
+		mCmdExecutor: tasks.MultiCmdExecutor,
+	}, true)
 }
--- a/tasks/infra/config/config_test.go
+++ b/tasks/infra/config/config_test.go
@@ -26,12 +26,7 @@ func TestRegisterWithCount(t *testing.T) {
 		registrationFunc func(tasks.Task, bool)
 	}
 
-	var expectedRegisteredTaskCount int
-	if runtime.GOOS == "windows" { //we don't register InfraConfigValidateJMX in windows due to bug
-		expectedRegisteredTaskCount = 6
-	} else {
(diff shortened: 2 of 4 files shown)


punkael (approved):
Looks great. Tested on Windows 10 VM.

PR #44, opened 2020-12-04
List available kafka brokers

uturunku1:
# Issue
No issue, this a new feature/health check for Infrastructure agent/Kafka OHI

# Goals
Our goal is to add a new task for our Infrastructure agent that will look into the Kafka integration configuration and validate that the setting `zookeeper_path` (found in kafka-config.yml) is a valid path. If this path allows us to connect to Kafka server/brokers, then the path is valid.

# Implementation Details
1. I added a new go file under the category: Infra/Env/ValidateZookeeperPath
2. You'll notice this file has the linux extension: ValidateZookeeperPath_linux.go. This task was built and tested with only linux environment in mind since this is the most popular environment for new relic customers. We see very few windows tickets coming in at least for Infra OHI.
3. As I was mentioning above, connecting to the list of brokers is the way we troubleshoot that the zookeeper_path is valid. To connect to that list of brokers, we are following these steps:
- Verify we have found the kafka-config.yml where this path is set.
- Parse the yml to collect the keys and values for this setting.
- put together a command line using the values found. The cmd goes like this: `/path-to/zookeeper-shell.sh localhost:2181 ls valueforzookeeper_path/brokers/ids`
- Additionally, we need to find out the path to zookeeper-shell, a script that helps us run the mentioned command line. We can find its path by looking at some env vars: $PATH, $KAFKA_HOME, $ZOOKEEPER_HOME. Any of those may give us a hint where that script is located.
- With all that info collect, we can run the cmdline. If it returns an error, then we'll attempt to run the same cmdline minus their zookeeper_path set. It is possible that the path the user set is invalid and we'll verify this if the cmdline succeeds without using the path set.

# How to Test
If you are interested in smoke testing this task, then you can:
1. Checkout into this specific branch
2. Build the go binary for linux: `GOOS=linux go build`
3. Grab the binary for the root directory(newrelic-diagnostics-cli) and bring into your linux box where you'll be testing the binary.
4. You do not need to have a kafka server running, but at least you'll need to install the new relic infra agent + the kafka integration: https://docs.newrelic.com/docs/integrations/host-integrations/host-integrations-list/kafka-monitoring-integration
5. Then place the go binary inside the /etc/newrelic-infra/ directory. You run the ValidateZookeeperPath task with this cmd: `newrelic-diagnostics-cli -y -t Infra/Env/ValidateZookeeperPath`. You'll probably get a failure result, which is to be expected, because you do no have a kafka server
6. Now if you want to get fancy and install/configure a kafka server, then I recommend this tutorial: https://linuxhint.com/install-apache-kafka-ubuntu/


tasks/infra/env/validateZookeeperPath_linux.go
```diff
@@ -0,0 +1,254 @@
...
+	if olderZkFileStatus.IsValid {
+		return binPath + "/zkCli.sh", true
+	}
+	return "", false
+}
+
+func (p InfraEnvValidateZookeeperPath) getKafkaBrokersList(zookeeperConfig ZookeeperConfig, zookeeperShellPath string, getArg string, brokersArg string) (bool, string) {
+	var hostPortArg string
+	if len(zookeeperConfig.Port) > 0 {
+		hostPortArg = "localhost:" + zookeeperConfig.Port
```
MichelLosier, line 183: Seems we ignore the zookeeper `host` provided in the infra integration config in favor of localhost, curious to the reason we favor that.

tasks/infra/env/validateZookeeperPath_linux.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+		//is not our job to validate those brokers IDs, just to make sure they are accessible
+		return true, fmt.Sprintf("We ran the command %s and succesfully connected to a broker or list of brokers:\n%s", cmd, string(cmdOutput))
+	}
+	//We found a path set in config file (Example, zookeeper_path: "/kafka-root"), so we'll append it to the brokers argument: zookeeper-shell.sh localhost:2181 ls /kafka-root/brokers/ids
+	customBrokersArg := zookeeperConfig.Path + brokersArg
+	customCmd := zookeeperShellPath + " " + hostPortArg + " " + getArg + " " + customBrokersArg
+	customCmdOutput, customCmdErr := p.cmdExec("/bin/bash", "-c", customCmd)
+
+	if customCmdErr != nil {
```
MichelLosier, line 207: nice fallback

MichelLosier (changes requested):
Looks good, just a few small comments/ questions! I do have a slight concern about dependencies on bash being on the system and the zookeeper script. Is the zookeeper cli script always installed with zookeeper?

tasks/infra/env/validateZookeeperPath_linux.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+func getZookeeperConfigFromYml(kafkaConfigPair *infraConfig.IntegrationFilePair) (ZookeeperConfig, error) {
+	zookeeperExtractedConfig := ZookeeperConfig{}
+	zookeeperPathBlobs := kafkaConfigPair.Configuration.ParsedResult.FindKey("zookeeper_path")
+
+	if len(zookeeperPathBlobs) > 1 {
+		return ZookeeperConfig{}, fmt.Errorf("Multiple keys found for zookeeper_path")
+	} else if len(zookeeperPathBlobs) == 1 {
+		zookeeperExtractedConfig.Path = zookeeperPathBlobs[0].Value()
```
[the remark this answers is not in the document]
uturunku1: ah yes! I actually did run into that index out of range error as I was smoke testing. I had comment out `zookeeper_path` in my kafka-config.yml. 
If `zookeeperPathBlobs` is empty, then we'll just ignore its value and `zookeeperExtractedConfig.Path` will have no value assign to it either. So when we get to the point when we want to use `zookeeperExtractedConfig.Path` value, this line should catch that we never got a value assigned to it: https://github.com/newrelic/newrelic-diagnostics-cli/blob/51b4b3ef2a13347288e1bcfab3bdfdd8dee513c3/tasks/infra/env/validateZookeeperPath_linux.go#L189

tasks/infra/env/validateZookeeperPath_linux.go
uturunku1: Good question! I was wondering the same thing about our kafka integration. The integration, as well as the TSEs when they troubleshoot, they completely ignore the host values and troubleshoot around localhost. So I copy the same behavior for nrdiag troubleshooting. I did do a little testing around this, using localhost vs another hostname when running ZK script tool in my command line and it gave me the same output for both. My guess is that somewhere in the ZK implementation this information may already be gathered implicitly or may not be as needed as other pieces of information, such as a port number.

uturunku1:
About your question `Is the zookeeper cli script always installed with zookeeper?`. Yes, it comes by default when you install zookeeper directly by itself or when you install Kafka that comes bundle with zookeeper and its ZK tools. The script itself, though, may have a different name/path based on the version installed.

MichelLosier (approved):
Looks good!

PR #45, opened 2020-12-15
Refactor java/jvm/permissions task

uturunku1:
# Issue

Addressing this Jira issue: https://newrelic.atlassian.net/browse/IHE-156
This task needed a major refactor for two reasons:
1. It was swallowing many errors that the user should be exposed to, and instead it's overwriting them with a vague message such as this one:

```
"Permission settings cannot be determined for the New Relic Agent JAR at location /app/weblogic/domains/mmCareLinkOneA/carelink/tools/newrelic.jarfor PID 12497. Problem is: The permissions settings for the Agent JAR are indeterminable"
```

when the real error is staying at debug level: https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/java/jvm/permissions_nix.go#L741

2. It has 4 major functions that are re-doing the work of finding values in config files, system properties, jvm args, env vars, and default places when we already have other tasks that are doing this job:
https://github.com/newrelic/newrelic-diagnostics-cli/blob/9fdd69492a63ceeee22c68d2ec83e3ca89851e6c/tasks/java/jvm/permissions_nix.go#L262
https://github.com/newrelic/newrelic-diagnostics-cli/blob/9fdd69492a63ceeee22c68d2ec83e3ca89851e6c/tasks/java/jvm/permissions_nix.go#L398
https://github.com/newrelic/newrelic-diagnostics-cli/blob/9fdd69492a63ceeee22c68d2ec83e3ca89851e6c/tasks/java/jvm/permissions_nix.go#L351
https://github.com/newrelic/newrelic-diagnostics-cli/blob/9fdd69492a63ceeee22c68d2ec83e3ca89851e6c/tasks/java/jvm/permissions_nix.go#L278

Not the author's fault. These other tasks did not exist when this task was built.

# Goals
1. Use other Diagnostics CLI tasks as dependencies to get the information about the logs, jar and temp directory locations to check for their permissions.
2. Bubble up the errors that were staying at log.Debug level.
3. Improve BaseLogCopy and logHelpers.go to capture the full paths and source of origin(env vars, jvm arguments, etc) of the new relic logs found. And then downstream that payload into Java/JVM/Permissions tasks

# Implementation Details
1. You may notice some small, unrelated changes. For example, changing the verbiage of the summary presented to the user when the appname.go displays a failure. As I was smoke testing in my Linux box(where I have infra agent installed), I noticed this task failing for Infra, which is not fair because appname task only applies to regular APM agents, so I made a change on the spot because otherwise, I may not do it later.
2. I did major changes to other files aside from Java/JVM/Permissions: Base/Log/Copy and Base/Log/LogHelpers.go

From my point of view, these changes were very much needed. That task (and its helper functions from LogHelpers) has access to information about all the new relic logs collected but most of that information is getting toss out. For example, the origins of those logs (if they come env vars versus config files, etc), and if they are secure files(meaning non-new relic logs). This information is not only relevant to Java/JVM/Permissions, which needs to know if a log comes from a system property or an env var so it can prioritize which log file to use to review its permissions, but it's also very relevant for TSE troubleshooting. If Diagnostics CLI collects multiple logs for, say, the Java agent, then how do they need which one to prioritize. If they see one of them comes from System Properties and the other comes from a default location, then they know the user is purposely using the one defined by System props.

3. You'll also notice I made changes to reportingTo_test.go and process_unix_test.go.  The first one got affected by changes made in BaseLogCopy and the second one got affected by a field I added to its struct. I needed that field to downstream to JavaJVMPermissions. So of course some of the tests break and I had to adjust them.

4. Finally JavaJVMPermissions.
- We'll be using a pointer to a struct called JavaAgentPermissions that will gather information about permissions provided to the :
agent Jar file
new relic log directory 
and the tmp directory (where the java agent temporarily drops API jar files)
- If at least one of them does not meet the agent permissions requirements, return a failure for the task
- if no failures, but at least one of them was not able to determine permissions, we'll return a warning for the task.

# How to Test
This task only runs on darwin and Linux. I recommend smoke testing this task by setting your permissions to readonly for your java app's directory(that should have the Java agent installed) and watch the task fail.
Instructions on how to install the Java agent: https://docs.newrelic.com/docs/agents/java-agent/installation/install-java-agent
Instructions on how to run this task in your java agent app:
- After cloning this repo, do `git checkout update-Java/JVM/Permissions`
- GOOS=linux go build (this will build the binary in the root directory for Diagnostics CLI
- Drop the binary into your java app root directory
- Make sure the application is running
- Run the task with: `./newrelic-diagnostics-cli -y -t Java/JVM/Permissions`


tasks/java/jvm/permissions_nix.go
```diff
@@ -76,674 +87,379 @@ func (p JavaJVMPermissions) Explain() string {
 /* Dependencies - Depends on the the SysPropCollect task to get PIDs with corresponding command-line args */
 func (p JavaJVMPermissions) Dependencies() []string {
 	return []string{
-		"Base/Env/CollectEnvVars",
```
punkael, line 79: I'm still getting wrong path for logs. Could have to do with this change in dependencies, where "Base/Log/Copy" may be returning a different path than "Base/Env/CollectEnvVars".

uturunku1: Update: After our discussion, it sounds like you were able to find out that wrong path for logs was a pre-existing bug coming from `main` branch.  And that this issue is no longer present in the changes made by this PR, wohoo🐎🏁.. Just as a reference to anyone who made see this comment in the future, the issue that Scotto had found that this Permissions task was reading comments from the yaml file, instead of ignore them, causing some unexpected results in the Permissions task. At this point, we believe that the library version used to parse yamls (version 2) was doing a little too much work when parsing config files. As of 3 months ago, we are now using version 3 of that library.

tasks/java/jvm/permissions_nix.go
```diff
@@ -76,674 +87,379 @@ func (p JavaJVMPermissions) Explain() string {
...
+		tempDir = os.TempDir() //may return a directory that does not exist but soon we'll find out as we check for this directory's permissions
+		tempDirSource = "the default tmp directory for the Operation System"
 	}
-	/* YAML file variable checks */
-	if newLogFileName.Value == "" {
+	//start assigning javaAgentPermissions values to temp directory
+	j.TempFilesCanCreate.Source = tempDirSource
+	j.TempFilesCanCreate.Value = tempDir
 
-		rawFileName, err := tasks.ReturnLastStringSubmatchInFile("log_file_name: (.+)", newYAMLFile.Value)
```
punkael, line 365: ReturnLastStringSubmatchInFile searches a file for any string match, regardless if it is a comment or not. It's removal solves for finding a remarked line in the yaml 🥳

uturunku1: You are so right! https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/fileSearch.go#L92
I wonder if the author of the task was aware of this effect.  I am guessing "no" because this is not how I would judge a current path to an existing log file. If a user comments out a key/value in a new relic config file, I would assume that the user is no longer storing the log file in that path.

punkael: Looks like this utility serves other functions, just used once off to grab a yaml key.

punkael (approved):

PR #46, opened 2020-12-20
Update infra/config/integrations match

uturunku1:
# Issue
The task Infra/Config/IntegrationsMatch incorrectly declares a failure for this validation if there is not a definition file for a specific integration installed.
A year ago, the infra agent updated their configuration to allow for no definition files:
https://docs.newrelic.com/docs/create-integrations/infrastructure-integrations-sdk/specifications/host-integrations-newer-configuration-format

# Goals
1.Update this task so it will not provide a failure status for an OHI integration that has no definition file and that has infra agent version 1.8.0 or higher
2. Use the payload from Infra/Agent/Version to check on the infra version
# Implementation Details

# How to Test
These changes were tested in a linux box that had the infra agent installed, version 1.13. It had the kafka, JMX and Docker OHIs configured. The docker-config. yml did not have a definition file, whereas kafka and JMX did.
The result was a success status, as expected.

### DIFF
--- a/tasks/infra/config/integrationsMatch.go
+++ b/tasks/infra/config/integrationsMatch.go
@@ -274,16 +274,17 @@ func validateIntegrationPairs(filePairs map[string]*IntegrationFilePair, isNewer
 			delete(filePairs, integration)
 			continue
 			// if pair is missing definition
-		} else if pair.Definition.Config == (config.ValidateElement{}.Config) && isNewerVersion {
-			//version 1.8.0 and higher do not require a definition file
-			continue
-		} else if pair.Definition.Config == (config.ValidateElement{}.Config) && !isNewerVersion {
-			matchErrors = append(matchErrors, IntegrationMatchError{
-				IntegrationFile: pair.Configuration,
-				Reason:          fmt.Sprintf("Configuration file '%s%s' does not have matching Definition file (%s-definition.yml)", pair.Configuration.Config.FilePath, pair.Configuration.Config.FileName, integration),
-			})
-			delete(filePairs, integration)
-			continue
+		} else if pair.Definition.Config == (config.ValidateElement{}.Config) {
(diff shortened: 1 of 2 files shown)


MichelLosier (changes requested):
Since the infra agent is backwards compatible to v3 integration config files, checking infra agent version alone may provide false negatives in the case where its a >= 1.8.0 agent using a integration configuration file v3 that still requires a definition file. 

I think we need here additionally a way to determine what is a v3 integration config file (requires definition file) and what is a v4 file. That way we can say if agent is >= 1.8.0 && config file is v4 then no definition file is required.

This docs section may provide some ideas on how we can distinguish those two: https://docs.newrelic.com/docs/create-integrations/infrastructure-integrations-sdk/specifications/host-integrations-newer-configuration-format#update

tasks/infra/config/integrationsMatch.go
```diff
@@ -297,6 +298,22 @@ func validateIntegrationPairs(filePairs map[string]*IntegrationFilePair, isNewer
...
+	/*
+		To update an older integration configuration to the new format, you must perform two steps. 1)Rename the "instances" top-level section to "integrations" 2)Remove the "integration_name" top-level section and add it to each integration entry (adding it is optional though)
+		https://docs.newrelic.com/docs/create-integrations/infrastructure-integrations-sdk/specifications/host-integrations-newer-configuration-format#update
+	*/
+	integrationsKeys := configElement.ParsedResult.FindKey("integrations")
+	/* Expect to find at least one. Example:
+		/integrations/0/name: nri-docker
+		/integrations/0/when/feature: docker_enabled */
+	instancesKeys := configElement.ParsedResult.FindKey("instances")
+	// Expect to find none
```
MichelLosier, line 311: great notes, thank you!

MichelLosier (approved):
Looks good, thank you!

MichelLosier (approved):
👍  2nd approval for merge conflict resolution

PR #52, opened 2021-01-12
Add rollback workflow

uturunku1:
# Issue
I had an issue with our latest release and luckily I was able to place a quick fix for this and re-run the release process. However, for situations where we do not have a quick fix and cannot roll forward but rather rollback, we would have to manually delete the previous release from Github, then delete the tag_name, then go use AWS CLI to delete (the artifacts that represent the current release version) and upload(the old artifacts that represent the previous version) from download.newrelic.com. Seems like one too many steps that could go wrong if someone has not memorized them and is in a rush.

# Goals
My goals are to automate the rollback process so if we do not have a hotfix release, then we can at least rollback safely and quickly.

# Implementation Details
- Adding a new workflow for Github actions
- creating a new script `rollback.sh` that gets called through the GH action.
- Notice that the aws cli commands that get run in the script are pointing to a testing s3 bucket. This is specifically for testing purposes and after we can verify that the workflow runs as expected, I'll make the changes to point to the real bucket.

# How to Test
The best way to test this rollback process is by deleting a fake release that I'll add for you here https://github.com/newrelic/newrelic-diagnostics-cli/releases/ . Please do not add a new release by yourself. When you are reading to test, let me know and I'll add the fake release. I do not want to add it ahead of time because this is a public space and I do not want visitors to notices a fake release sitting there ;)

Once you delete the fake release, this will trigger the rollback process and you'll be able to see it in action here:
https://github.com/newrelic/newrelic-diagnostics-cli/actions?query=workflow%3ARelease
You can also witness the changes in our downloads page: https://download.newrelic.com/nrdiag/test/
If you visit this site right now, you should see this file:
nrdiag_1.12.32.zip
nrdiag_1.11.31.zip
nrdiag_latest.zip
version.text
After running the rollback process, you should see:
nrdiag_1.11.31.zip
nrdiag_latest.zip
version.text

Additionally, this PR will need to be approved before it can get tested. Why? Because the only way to trigger the workflow is by deleting a release from this interface: https://github.com/newrelic/newrelic-diagnostics-cli/releases/
If you were to delete a release without merging the changes, then nothing would get triggered because the changes only live in this branch.
It is safe to merge the changes because right now they are pointing to a testing bucket so nothing will get affected within the "official" bucket.


pmillspdx (approved):
Looks good to me. Ready to test it once it's merged.

PR #54, opened 2021-01-18
bubble up error when we are unable to get dotnet core version

uturunku1:
# Issue
1. The task DotnetCoreEnvVersions is not showing the errors when it is unable to find the .NET core SDK version. The error summary only says: "Error determining the version .NET Core installed" when we actually do have access to the error itself. So we just need to add it to the summary.
2. The payload of DotnetCoreEnvVersions (a slice of strings. Each string is supposed to be a version) is used as an upstream payload in the task DotNetCore/Requirements/DotNetCoreVersion. When iterating through that upstream payload, we were attempting to get major, minor digits for each item (`coreVersion`): `majorVer, minorVer, _, _ := tasks.GetVersionSplit(coreVersion)` . But what if the item we are iterating through is not a string that has well-defined digits we can parse through? Ex: `.NET Core 5.0.101` instead of just `5.0.101`

# Goals
1. Add errors output to the result summary message if we were unable to get a NET core SDK version in DotnetCoreEnvVersions task
2. Take into account that the payload received from DotnetCoreEnvVersions could contain a string that is not a parseable version.

# Implementation Details

# How to Test

### DIFF
--- a/tasks/dotnetcore/env/versions.go
+++ b/tasks/dotnetcore/env/versions.go
@@ -1,13 +1,13 @@
 package env
 
 import (
+	"fmt"
 	"errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
 	"runtime"
-	"strconv"
 	"strings"
 
 	"github.com/newrelic/newrelic-diagnostics-cli/logger"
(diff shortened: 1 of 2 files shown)


KayLotura2 (approved):
Looks solid to me

PR #55, opened 2021-01-20
Update base/Log/Copy task for .NET agent logs

uturunku1:
# Issue

1)the task Base/Log/Copy can collect a lot of instances of .NET profiler logs from standard locations:
```JSON
{
   "FileName": "NewRelic.Profiler.7748.log",
   "FilePath": "C:\\ProgramData\\New Relic\\.NET Agent\\Logs\\",
   "Source": {
			"FoundBy": "Found by looking at standard locations",
			"KeyVals": null,
			"FullPath": "C:\\ProgramData\\New Relic\\.NET Agent\\Logs\\NewRelic.Profiler.7748.log"
			},
   "IsSecureLocation": false,
   "CanCollect": true,
   "ReasonToNotCollect": ""
}
```
Here I am just showing one instance from a list of 243 found in the nrdiag-output.json
The issue here is that with such a long list, the size of nrdiag-output.json file can grow really quickly and exceeded our size limitation placed in Zendesk. If the JSON file is too large, then it won't get upload into our nrdiag sidebar app (which is used by TSEs for easy readability of this file).

2)We are not collecting NET logs that have been configured through env vars

# Goals

1. Decrease the number of profiler logs we collect by limiting the number of days we collect from. Only collect from the last 4 days.
2. Move the logic for collecting only recent files(no older than 4 days in the case of profiler and 7 days for other logs) to earlier in the process so it is only applicable to when we are finding files in `standard locations`, and not applicable to files that were found by looking into the user's settings. 
Right now we are running this logic too late in the process, once we found logs from multiple sources. However, the risk for collecting old, ***unwanted*** log files only comes from searching files in `standard locations` because this location does not necessarily represent the location where the user intended to configured their log file creation. 
3. To not display all the NET profiler logs in the payload, only one sample. 
4. Include .NET agent log env var when we are looking for log paths that have defined through env vars

# Implementation Details

**Changes related to downsizing the quantity number of NET profiler logs we collect and display in the payload:**
1. In Base/Log/logHelpers.go: I created 2 new variables `defaultMaxNumDays = 7` and `profilerMaxNumDays = 4` to predetermined that for "recent" .NET profiler logs, we should only collect logs from the last 4 days. For any other logs, "recent" represents last 7 days.
Those variables get used in `func determineFilesDate`. So now that function gets call in the only 2 places where I see risk of collecting old files: `func getLogPathsFromCurrentDirOrNamePatters` and `func getLogPathsFromStandardLocations`. What those two functions have in common is that they look for files patterns(regex) instead of specific file names. The risk here is that with a pattern name, we can end up matching multiple file names and inevitability some of them will older than others.

2. In Base/Log/Copy: Once we have verified that we have found valid log file paths to new relic logs, I am going to check if, among those valid log file paths, we have paths that point to .NET log profilers. If we do, then I am only going to select one profiler to be added to the payload result.  At this point, we have already captured all profilers into the Result.FilesToCopy field, meaning that TSEs will still be able to find all the profiler logs zipped into the nrdiag-output.zip

**Changes related to adding env vars for NET agent logs:**
1. In Base/Log/logHelpers.go: I added `NEWRELIC_PROFILER_LOG_DIRECTORY` and added a comment around `NEW_RELIC_LOG` to clarify this is also an env var used by the NET agent (This is a recent change: https://docs.newrelic.com/docs/release-notes/agent-release-notes/net-release-notes/net-agent-83700)
2. `NEWRELIC_PROFILER_LOG_DIRECTORY`points to a path directory, unlike other env vars that pointed to a filename or that are a full path that includes the filename. Because of this, I had to refactor `func getLogPathFromEnvVar` to verify if the path is a path to a directory

```
if pathInfo.IsDir() {
		unmatchedDirKeyToVal[logEnvVar] = logPath
		return LogElement{}, true
}
```

3. I also added a new function `func getLogPathFromUnmatchedKeys`. As I was smoke testing the new functionality to look for `NEWRELIC_PROFILER_LOG_DIRECTORY` and `NEW_RELIC_LOG` env vars,  I noticed that these 2 env vars do not complement each other, even though one point to a directory and the other points to the filename. The `NEW_RELIC_LOG` does not take the filename for profiler logs but for the Agent's log file. The Base/Log/Copy task failed at finding the profiler logs because it attempted to combine the values for `NEWRELIC_PROFILER_LOG_DIRECTORY` and `NEW_RELIC_LOG`. To fix this, I created `func getLogPathFromUnmatchedKeys` which looks for "unmatched" keys, meaning keys that need a pair key to get the full path to a log file. `NEWRELIC_PROFILER_LOG_DIRECTORY` and `NEW_RELIC_LOG` are examples of unmatched keys. `func getLogPathFromUnmatchedKeys` will attempt to pair them, but if it fails then it attempts to combine the values of those unmatched keys with `standard locations` and `standard logfile name`. In this case, it will combine the value of `NEWRELIC_PROFILER_LOG_DIRECTORY` with the standard log name `NewRelic[.]Profiler.*[.]log$`
4. One last refactor that I included in this PR is the new function `func setLogElement`. This function's purpose is to add values, in a more concise way, to the many fields the LogElement struct has :

```
LogElement{
		FileName:           filename,
		FilePath:           dir,
		Source:             LogSourceData{
	                                      FoundBy  string
	                                      KeyVals  map[string]string
	                                      FullPath string

                 },
		IsSecureLocation:   isSecureLocation,
		CanCollect:         canCollect,
		ReasonToNotCollect: reasonCannotCollect,
	}
```
Before we had to create this giant struct instance in multiple places. Now we just call the helper function setLogElement to set the values for those fields and the number of code lines has reduced quite a bit.

# How to Test


punkael (approved):
Looks good. Thanks for the in-depth description, it was super helpful for the review!

PR #56, opened 2021-01-22
Check for yarn

punkael:
# Issue

# Goals

# Implementation Details

# How to Test


tasks/node/env/dependencies.go
```diff
@@ -34,6 +34,24 @@ func (p NodeEnvDependencies) Dependencies() []string {
...
+	// 	}
+	// }
+
+	// if packageManager == "" {
+	// 	return tasks.Result{
+	// 		Status:  tasks.None, //tasks.Warning?
+	// 		Summary: "This task did not because we could not found npm or yarn installed",
+	// 	}
+	// }
+
```
cade-conklin, line 54: If these aren't needed, the comments can be deleted

cade-conklin (changes requested):

cade-conklin:
@punkael would you mind pulling in the latest changes from `develop`?

daffinito:
I updated the yarn task to use dependency injection and added tests

daffinito:
Updated the dependencies task to check for yarn if npm isn't installed.

PR #57, opened 2021-01-22
updated openJDK from 7.14 to 7.15. Updated HotSpot from 7.14 to 7.15

brianyangNR:
# Issue
Versions were not up to date
# Goals

# Implementation Details

# How to Test


uturunku1 (approved):
oh, yes! The java agent keeps expanding its supportability for JVM vendors and versions: https://docs.newrelic.com/docs/agents/java-agent/getting-started/compatibility-requirements-java-agent#jvm
So I was not surprised to see that we need to update our supportability as well. Thanks for making the necessary update!

PR #60, opened 2021-01-25
reformat ValidateLicenseKey summary

uturunku1:
# Issue
Just fixing the result summary for this task because I noticed that it was a little difficult to read. It needed some new lines in between words like this:

```
The license key found in\n%s\nhas a valid New Relic 
```

And it needed less spacing at the bottom:
from `reach out to New Relic Support.\n\n` to `reach out to New Relic Support.\n`

# Goals
1) Reformat the summary result to be a bit more readable
2) Adjust unit test that was failing because of the change in the summary format.

# Implementation Details

# How to Test


KayLotura2 (approved):
Looks clean and simple enough to me. I'm no expert in Go yet, but I assume these are best practices?

uturunku1:
No necessarily Go related best practices, I just saw this little reformat as a need for how we display results in the terminal UI for users. I thought the result summary sentences were looking a little too smashed together at the beginning of the summary, and then a little too much empty space at the end of the summary. That is why a added a new line at the beginning of the message and remove an empty line at the end of it.

PR #63, opened 2021-01-27
Node version module update (PR on different branch)

KayLotura2:
# Issue
Bug with using file scraping to find `newrelic` Node Agent Module Version.
Updated to use already implemented payload.

# Goals
Cuts down extra work, returns more reliably accurate info.

# Implementation Details
Small changes, unit tests.

# How to Test
N/A, unit tests cover the small change.


tasks/node/agent/version.go
```diff
@@ -36,63 +38,46 @@ func (t NodeAgentVersion) Dependencies() []string {
...
-		return result
-	}
-	logs, ok := upstream["Base/Log/Copy"].Payload.([]logtask.LogElement)
-	if !ok {
-		result.Status = tasks.None
-		result.Summary = "Node agent version not detected"
-		return result
+		return tasks.Result{
+			Status:  tasks.None,
+			Summary: "Node agent not detected",
```
uturunku1, line 47: Here it should read: "Node Agent not detected. This task did not run"

tasks/node/agent/version.go
```diff
@@ -36,63 +38,46 @@ func (t NodeAgentVersion) Dependencies() []string {
...
-			result.Status = tasks.Info
-			result.Summary = nodeLog.AgentVersion
-			result.Payload = nodeLog.AgentVersion
-			return result
+	if upstream["Node/Env/Dependencies"].Status == tasks.Info {
+		nodeModuleVersions, ok = upstream["Node/Env/Dependencies"].Payload.([]NodeEnv.NodeModuleVersion)
+		if !ok {
+			return tasks.Result{
+				Status:  tasks.None,
+				Summary: "Node module version not detected",
```
uturunku1, line 55: Similar here: "Node Modules not detected. This task did not run"

tasks/node/agent/version.go
```diff
@@ -36,63 +38,46 @@ func (t NodeAgentVersion) Dependencies() []string {
...
-			logElementNode.MatchFound = false
-		} else {
-			logElementNode.Logfile = logElement.FilePath + logElement.FileName
-			logElementNode.AgentVersion = logAgentVersion
-			logElementNode.MatchFound = true
+	agentVersion := getNodeVerFromPayload(nodeModuleVersions)
+	if agentVersion == "" {
+		return tasks.Result{
+			Status:  tasks.Warning,
+			Summary: "Node Agent Module not found for newrelic",
```
uturunku1, line 64: Maybe we should add a more explicit message to the user, like:  "We were unable to find the 'newrelic' module required for the Node Agent installation. Make sure to run 'npm install newrelic' and verify that 'newrelic' is listed in your package.json"

uturunku1 (changes requested):
Loooking goood! I requested a few changes. I also noticed that we still have the `logNodeAgentVersion` struct defined at the top of the file version.go. If we are not using that struct anywhere anymore, let's get rid of it.

tasks/node/agent/version.go
KayLotura2: That's a great message. Thanks for the suggestion.

tasks/node/agent/version.go
```diff
@@ -29,70 +25,51 @@ func (t NodeAgentVersion) Explain() string {
 // Dependencies - Returns the dependencies for this task.
 func (t NodeAgentVersion) Dependencies() []string {
 	return []string{
-		"Node/Config/Agent",
-		"Base/Log/Copy",
+		"Node/Env/Dependencies",
```
uturunku1, line 28: We were missing listing this relevant dependency prior to verifying its status and payload.

tasks/node/agent/version.go
```diff
@@ -29,70 +25,51 @@ func (t NodeAgentVersion) Explain() string {
...
 
 // Execute - The core work within this task
 func (t NodeAgentVersion) Execute(options tasks.Options, upstream map[string]tasks.Result) tasks.Result {
-	var result tasks.Result
-
-	if upstream["Node/Config/Agent"].Status != tasks.Success {
-		result.Status = tasks.None
-		result.Summary = "Node agent not detected"
-		return result
+	if upstream["Node/Env/Dependencies"].Status != tasks.Info {
```
uturunku1, line 34: I removed our check for Node/Config/Agent because Node/Env/Dependencies already checks that Node/Config/Agent ran before it.

tasks/node/agent/version.go
```diff
@@ -29,70 +25,51 @@ func (t NodeAgentVersion) Explain() string {
...
-	searchString := `Node[.]js[.] Agent version:\s*([0-9.]+);`
-	filepath := logElement.FilePath + logElement.FileName
-	agentVersion, err := tasks.ReturnLastStringSubmatchInFile(searchString, filepath)
-	if err != nil {
-		return "", err
+	log.Debug("Agent version", agentVersion)
+	return tasks.Result{
+		Status:  tasks.Info,
+		Summary: fmt.Sprintf("Node Agent Version %s found", agentVersion),
+		Payload: agentVersion,
```
uturunku1, line 62: We were missing returning a Payload and while I was smoke testing it became obvious because other tasks rely on this payload, and of course, they were not running or failing.

tasks/node/agent/version.go
```diff
@@ -29,70 +25,51 @@ func (t NodeAgentVersion) Explain() string {
...
+}
 
-	if len(agentVersion) > 0 {
-		// we've got a match, return the first capture group
-		return agentVersion[1], nil
+func getNodeVerFromPayload(payload []NodeEnv.NodeModuleVersion) string {
+	for _, nodeModuleVersion := range payload {
+		if nodeModuleVersion.Module == "newrelic" {
+			return nodeModuleVersion.Version
+		}
```
uturunku1, line 70: We want to return only `nodeModuleVersion.Version` instead of `nodeModuleVersion.Module+ nodeModuleVersion.Version` because the downstream tasks that will use this information/payload only expect a string made out of digits.

tasks/node/agent/version_test.go
```diff
@@ -15,132 +16,130 @@ func TestNodeAgentVersion(t *testing.T) {
...
 		})
-
-		Context("When given a logElement with a malformed filepath", func() {
+		Context("When given a Node Agent Module is found", func() {
 			BeforeEach(func() {
-				incomingLogs = []logtask.LogElement{
-					logtask.LogElement{
-						FileName: "newrelic_agent.BADFILE.log",
-						FilePath: "../../fixtures/node/",
+				upstream = map[string]tasks.Result{
```
uturunku1, line 112: I added one additional test for a successful case when we do get a node agent version

tasks/node/env/dependencies.go
```diff
@@ -44,7 +44,7 @@ func (p NodeEnvDependencies) Execute(option tasks.Options, upstream map[string]t
 	if upstream["Node/Config/Agent"].Status != tasks.Success {
 		return tasks.Result{
 			Status:  tasks.None,
-			Summary: "Node agent not detected. This task did not run",
+			Summary: "Node agent config file not detected. This task did not run",
```
uturunku1, line 47: The following small changes are completely unrelated to this PR. It is a very nitpicky change in the summary of other tasks to make it more explicit that they did not run because we did not find a config file. The previous summary seems a little less clear to me.

uturunku1 (approved):

PR #73, opened 2021-02-04
Update task.Error summaries for type assertion error

uturunku1:
# Issue
`tasks.Error` status signifies that nrdiag could not finish running a task because we ran into an unexpected error. So something wrong went on our end and it is not the customer's fault. 

We have a few of those `tasks.Error` throughout our code, especially for when we run into a "type assertion error", is a Golang type-specific kind of error. 

If a customer gets a `task.Error` status occasionally they will report it to TSEs casually as they are working through a Support ticket, and then sometimes TSEs will report it to us, sometimes not. 

Ideally, we would always get reported on this kind of error because this is something that we need to immediately fix on our end.

# Goals

Now that we have a public space to report issues through Githu issues, we want to standardize a lot of our `tasks.Error` summaries to direct customers to our Github issues or even discuss.newrelic.com to report this error to us. 

# Implementation Details
1) added these two global variables that contain the message will give users when they ran into a Type assertion Error
https://github.com/newrelic/newrelic-diagnostics-cli/blob/7eeb11aa4caa8f63a13d3590aebdf30c0a4b3dec/tasks/tasks.go#L30
https://github.com/newrelic/newrelic-diagnostics-cli/blob/7eeb11aa4caa8f63a13d3590aebdf30c0a4b3dec/tasks/tasks.go#L27
These variables will get called everywhere where we use to have a result summary that indicated we ran into a type assertion error or any other place where did not have a result summary but instead we were just logging the error and hiding the error to ourselves.

2) I was making these changes I noticed in a few tasks we are attempting to collect an upstream task.Payload without checking first if the upstream task succeeding at getting a Payload: https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/ruby/config/agent.go#L44
This may have seen as an acceptable practice before because if that code line failed with a type assertion error, we would not notice because we have a backup upstream option right afterwards: https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/ruby/config/agent.go#L61 . The error would just get logged.
However, now that we are returning the type assertion error rather than just logging it, then without a check to make sure we have a payload, then the task will possibly exist early because of a type assertion error. That is why I added this check before attempting to access the payload: https://github.com/newrelic/newrelic-diagnostics-cli/blob/7eeb11aa4caa8f63a13d3590aebdf30c0a4b3dec/tasks/ruby/config/agent.go#L44

3) A minor change that I included in this PR, is to add the missing payload in some of the tasks that detect the existence of an agent. For example Ruby/Config/Agent, Node/Config/Agent and Python/Config/Agent.
The Go lint library warned that these tasks we had a slice variable that was going unused: https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/ruby/config/agent.go#L74 
So we had forgotten to add them to the payload: https://github.com/newrelic/newrelic-diagnostics-cli/blob/7eeb11aa4caa8f63a13d3590aebdf30c0a4b3dec/tasks/ruby/config/agent.go#L59
This was a small unnoticed issue because so far we have not used any of the payloads coming from these tasks in other downstream tasks.
They have now been added to the payload:
https://github.com/newrelic/newrelic-diagnostics-cli/blob/7eeb11aa4caa8f63a13d3590aebdf30c0a4b3dec/tasks/tasks.go#L73

# How to Test
Build the binary against this branch with `go build`
Then bring the binary to an app directory. Any application that has a new relic agent installed will work.
The run the binary without any flags so it will attempt to randomly troubleshoot for any agent: `./newrelic-diagnostics-cli`


tasks/dotnet/config/agent_windows.go
```diff
@@ -44,41 +44,46 @@ var webAppConfigKeys = []string{
...
+
+	// validate the config files elements
+	filesToAdd, err := checkConfigs(configFiles) //err is a boolean
+	if err {
+		return tasks.Result{
+			Status:  tasks.Warning,
+			Summary: "Unable to validate the .NET agent config files because the files do not contain typical .NET agent configuration settings.",
+		}
+	}
+	// no error means at least one file validated
```
KayLotura2, line 81: I super appreciate these doc strings

KayLotura2 (approved):
I assume `go.mod` and `go.sum` are go specific information files, maybe for testing purposes? Is that right? 

When building a struct in go, it looks like, at least with this repo, we try to have key value pairs spaced such that all values are aligned in a column? I like it, makes it easy to read values quickly. 

This looks pretty comprehensive, and it smoke tests fine in my personal repo.
Thanks for all the detail handling! This is really great.

uturunku1:
oh yes! thanks for asking about the changes made on go.mod and go.sum. I forgot to mention those in my description. 
These files are how Golang manages its dependencies. It's very cool because it self-updates as soon as I run a go command, for example `go test`. This is a good explanation I found for how `go.mod` works: "The go command resolves imports by using the specific dependency module versions listed in go.mod. When it encounters an import of a package not provided by any module in go.mod, the go command automatically looks up the module containing that package and adds it to go.mod, using the latest version. (“Latest” is defined as the latest tagged stable (non-prerelease) version, or else the latest tagged prerelease version, or else the latest untagged version.) In our example, `go test` resolved the new import rsc.io/quote to the module rsc.io/quote v1.5.2. It also downloaded two dependencies used by rsc.io/quote, namely rsc.io/sampler and golang.org/x/text. Only direct dependencies are recorded in the go.mod file"
As for `go.sum`, this was Go official documentation says about those crazy hashes you may have noticed in that file:
 "The go.sum file contains cryptographic hashes of the module's direct and indirect dependencies. When the go command downloads a module .mod or .zip file into the module cache, it computes a hash and checks that the hash matches the corresponding hash in the main module's go.sum file. go.sum may be empty or absent if the module has no dependencies or if all dependencies are replaced with local directories using replace directives." @KayLotura2

PR #74, opened 2021-02-06
update some Node agent tasks

uturunku1:
# Issue
As I study new tickets where nrdiag task fails to do its job, then I do my best to update the task so it will not fail in the future.

Recently I found a Node agent support ticket(https://newrelic.zendesk.com/agent/tickets/441655) where nrdiag failed for two tasks:

1) Base/Config/Collect: it should have thrown a warning or failure status because we did not collect a config file for the node agent. The reason we didn't notice this was a problem is because we did not take into account the env var users can configure to not use a config file: `NEW_RELIC_NO_CONFIG_FILE`.

2)Node/Config/Agent: We did not run many Node agent tasks they will only if Node/Config/Agent succeeds. Node/Config/Agent fails when it does not find a config file, but that is not a fair way to judge if a user has configured the node agent because they can skip using a config file, and configure through env vars.

# Goals
1) Account for `NEW_RELIC_NO_CONFIG_FILE` in Base/Config/Collect and warn the customer that we were unable to collect a config file because they set this env var to true. 

2) Also, I seize the opportunity that I was working on this task to make an unrelated but needed change. 
I am adding a check for when users say "no" to nrdiag collecting secure config files: https://github.com/newrelic/newrelic-diagnostics-cli/blob/cc7a3a882df4b1f98c1c30565901fcc64fa73639/tasks/base/config/collect.go#L201
This information can be relevant for TSE to understand that nrdiag did not fail at collecting a secure config file, but it was the user that reject us collecting the config file.

3) Updating `Node/Config/Agent` to not rely on the presence of a config file but also look for the following env vars to know that the user is attempting to configure the node agent: https://github.com/newrelic/newrelic-diagnostics-cli/blob/cc7a3a882df4b1f98c1c30565901fcc64fa73639/tasks/node/config/agent.go#L23 . These are the minimum env vars user will need to configure, any other env vars are optional.

I consulted with the Node agent engineers to make sure that the env var `NEW_RELIC_NO_CONFIG_FILE` is only used by the Node agent. If it was used by other agents then this task will mistakenly assume that we are in a node environment when in fact we are in another agent environment that happens to use the same env vars.

# Implementation Details

# How to Test
This task should smoke tested by building the binary `go build` with this branch and then dropping the binary in a node application that has the node agent installed. The node agent should not use a config file but use the env vars mentioned here https://github.com/newrelic/newrelic-diagnostics-cli/blob/cc7a3a882df4b1f98c1c30565901fcc64fa73639/tasks/node/config/agent.go#L23  
For more details on how to configure the Node agent: https://docs.newrelic.com/docs/agents/nodejs-agent/installation-configuration/nodejs-agent-configuration 
Then run the nrdiag node tasks with the flag `suites`: ./newrelic-diagnostics-cli -suites node

### DIFF
--- a/tasks/base/config/collect.go
+++ b/tasks/base/config/collect.go
@@ -26,6 +26,8 @@ var configEnvVarKeys = []string{
 	// PHP agent does not support config env vars
 }
 
+var noConfigEnvVar = "NEW_RELIC_NO_CONFIG_FILE"
+
 var patterns = []string{
 	"newrelic[.]yml",
 	"newrelic[.]xml",
@@ -160,7 +162,7 @@ func (p BaseConfigCollect) Execute(options tasks.Options, upstream map[string]ta
 	//Find insecure paths
 	foundSecureConfigs := tasks.FindFiles(secureFilePatterns, paths)
 
-	var invalidConfigFiles []string
+	var invalidConfigFiles, cannotCollectConfigFiles []string//will represent the secure files that the user reject nrdiag to collect at the prompt
 	var warningSummaryOnInvalidFiles string
 
 	for _, secureConfig := range foundSecureConfigs {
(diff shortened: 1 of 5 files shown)


KayLotura2 (approved):
This looks solid to me, pulled it down, and prodded it a bit. 
Noticed we have an integration test failing, but that's an issue on main, so this PR should be good to go!

PR #77, opened 2021-02-14
Update javagent

uturunku1:
# Issue
1) The Java/Env/Process task targetting Windows OS is out of date with the Java/Env/Process task for Linux and Darwin. The main difference between those two tasks is that the Windows one looks for the newrelic jar file by searching specifically for a file named "newrelic.jar". The other task is more flexible in its search and it will allow for a file named newrelic-4.6.0.jar, which is good because some new relic jar files can include the version number in the name.

2) The Java/Config/Agent looks for a "newrelic.jar" file but, like mentioned above, current versions of the java agent do not require this specific name for the JAR file

# Goals
1) To update the Java/Env/Process windows task to match the Linux and Darwin Java/EnvProcess task
2) To update Java/Config/Agent to look in the current directory for any JAR file that has "newrelic" name included in it.

# Implementation Details
1) Java/Env/Process
Originally I was planning to update this file https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/java/env/process_windows.go to look exactly like this file if possible https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/java/env/process.go
But then I asked myself, what is the point of making 2 copies of the same thing? I might as well delete process_windows.go and allow process.go to run for all OS.  

Most of the operation ran in `process_windows.go` are similar to the ones seen in `process.go`, except that `process_windows.go` uses this helper function `GetProcInfoByName` (https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/java/env/process_windows.go#L59) that is only defined as a windows helper function (https://github.com/newrelic/newrelic-diagnostics-cli/blob/cbb28332cb7efcf23154a2963aadab9a06d9dded/tasks/taskHelpers_windows.go#L200)

That is probably why the engineers that built this task opted for creating different files targetting different operation systems rather than creating one single `process.go` that runs by default in all 3 OS.

I smoke tested this helper function `GetProcInfoByName` to see what kind of data is surfacing that `process.go` cannot surface right now. 

I had 2 JVM processes running at the time and logging the output of `GetProcInfoByName` showed me this:

```
[{java.exe "C:\Program Files (x86)\Common Files\Oracle\Java\javapath\java.exe" -jar -javaagent:newrelic\newrelic.jar build\libs\lucessqs-1.0-SNAPSHOT.jar C:\Program Files (x86)\Common Files\Oracle\Java\javapath\java.exe 616} 
{java.exe java  -jar "C:\Program Files\New Relic\nrjmx\nrjmx.jar" --hostname localhost --port 9010 --username admin --password admin C:\Program Files (x86)\Common Files\Oracle\Java\javapath\java.exe 5724}]
```

This is not much different from just iterating through the Java processes IDs and getting their cmd line arguments, which is what `process.go` does (https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/java/env/process.go#L86)

This is how those command-line arguments look like for the same 2 processes when you are iterating through the processes IDs:

```
"C:\Program Files (x86)\Common Files\Oracle\Java\javapath\java.exe" -jar -javaagent:newrelic\newrelic.jar build\libs\lucessqs-1.0-SNAPSHOT.jar

java  -jar "C:\Program Files\New Relic\nrjmx\nrjmx.jar" --hostname localhost --port 9010 --username admin --password admin
```

Though there maybe just a tad more granularity in what `GetProcInfoByName` can provide, the results for both approaches will be the same because the task will still be focused on finding the -javaagent flag among these args:

```
"CmdLineArgs": [
								"\"C:\\Program",
								"Files",
								"(x86)\\Common",
								"Files\\Oracle\\Java\\javapath\\java.exe\"",
								"-jar",
								"-javaagent:newrelic\\newrelic.jar",
								"build\\libs\\lucessqs-1.0-SNAPSHOT.jar"
				],
```

So in the end I opted out of getting rid of `process_windows.go` and having `process.go` enabled in Windows.

2) Java/Config/Agent
Expand the logic of `checkForJar() ` helper function to check in the current directory for either a file that matches `(newrelic)?i([\S]+)?\.jar` or a directory that matches `(newrelic)?i`. In the latter case, if we found a match, we'll aditionally look into that directory for a jar file that matches `(newrelic)?i([\S]+)?\.jar`.

# How to Test
These changes can be tested in a Windows VM that has a Java agent application installed.
For instructions on how to install the java agent:  https://docs.newrelic.com/docs/agents/java-agent/installation/install-java-agent
The minimum that you'll need is to have the newrelic.yml config file in the same directory where your java app lives and the newrelic.jar file. The java app you'll be testing with can be as simple as a Hello World kind of app. Some examples here:
https://tomcat.apache.org/tomcat-7.0-doc/appdev/sample/
https://www.michael-thomas.com/tech/apache/tomcat/tutorial_firststeps_tomcat/index.htm

After you have this setup, build the Go binary against this branch for Windows OS: `GOOS=windows go build`
And bring the binary to your java app. Run all the nrdiag tasks for Java with the cmd: `newrelic-diagnostics-cli -y -suites java`
The `-y` option is to say yes to all prompts.
If something is not correct in your app setup or you did not configure the java agent correctly in your app, I am hoping nrdiag will warn you about those issues. So it will be like a real troubleshooting experience our customers would go through.


tasks/java/config/agent.go
```diff
@@ -185,17 +186,57 @@ func checkConfig(configs []config.ConfigElement) ([]config.ConfigElement, bool)
...
-	jarNames := []string{
-		"newrelic.jar",
+	//check for existence of this specific file name in the user's system. We do not look for a regex/pattern name because it would be a never ending search
+	if tasks.FileExists("newrelic.jar") {
+		log.Debug("Jar file found, setting true")
+		return true
+	}
+	//Now we can attempt to find a newrelic filename pattern, but only in the current directory
+	jarRgx := regexp.MustCompile(`(newrelic)?i([\S]+)?\.jar`)
+	dirRgx := regexp.MustCompile(`(newrelic)?i`) //in case they put the jar inside a newrelic directory
```
KayLotura2, line 196: I'm not certain the ?i portion of the regex is doing what we want here.
It seems to be allowing for `'newrelic' or 'any string with an i in it'`
Playing around with it here: https://golang.org/pkg/regexp/

It looks like adding parenthesis around (?i) if the idea is that the `.jar` is meant to be case insensitive
And moving (?i) to the front of `newrelic` if we want that to be case insenstitive.

Code snippet that I've been testing on: 
```
package main

import (
	"fmt"
	"regexp"
)

func main() {
	// Compile the expression once, usually at init time.
	// Use raw strings to avoid having to quote the backslashes.
	var validID = regexp.MustCompile(`(i?)(newrelic)([\S]+)?\.jar`)

	fmt.Println(validID.MatchString("bluerelxc.jar"))
	fmt.Println(validID.MatchString("bluerelic.jar"))
	fmt.Println(validID.MatchString("newrelic.jar"))
	fmt.Println(validID.MatchString("newRelic.jar"))
	fmt.Println(validID.MatchString("nerwRelic.JAR"))
}
```

tasks/java/env/process_windows.go
```diff
@@ -1,97 +0,0 @@
...
-		//	log.Debug("Error getting command line options")
-		//}
-		/* check to make sure the command line args contain the new relic Java Agent JAR
-		change the result status and summary if one or more processes with newrelic.jar are found
-		this check assumed the java agent jar is called newrelic.jar, but this is not a requirement with more recent agent versions */
-
-		if strings.Contains(value.CommandLine, "newrelic.jar") {
-			procObj := procObjs[0]
-			for _, proc := range procObjs {
-				if proc.Pid == int32(value.ProcessId) {
```
KayLotura2, line 80: Int32, will that be necessary for processing windows process IDs in the new java agent process?
I suspect not because we are comparing strings rather than numeric IDs but I'm uncertain.
This may not be a thing.

PR #87, opened 2021-07-28
Attachment key option

cade-conklin:
# Issue
We want to have an option for a user to auto-upload the Diagnostics CLI Output without using an attachment key.  Currently, they must supply a valid attachment key after the -a flag (./nr_diag_x64 -a <insert_attachment_key>).  The reason we want to change this functionality is so that the Diagnostics Nerdlet can gleam CLI runs from an account drop down.  Right now, there is no way to upload the Diagnostics CLI Output without an attachment key.  For flags that take in strings as parameters, there is no way to make the parameters optional.

# Goals
We want to give the user the option to include an attachment key or omit one when using our option flags.  We also want to be able to use a valid license key from a user's Diagnostics CLI run to upload to our s3 bucket and haberdasher so that we can access that output by account in the Diagnostics Nerdlet. In order to do that, we want the CLI to be able to send a/some valid license key(s) to the Haberdasher endpoint in order to get an upload URL.  We want to be able to make the processes for auto-attach and attachment key separate so the user can do both during a run.

# Implementation Details
In order to give the user an option to auto-attach their output and attach their output to a ticket, they will now have two different flags to do so.  -a and -attach will automatically upload output by any valid license keys found during Diagnostics CLI runs.  -ak <string> and -attachment-key <string> will upload by a valid license key parameter given by the user.  When processing the uploads, the CLI will look at the two different upload options and will determine if an upload with an attachment key or upload with a license key (if either exist) is necessary.  The valid license key(s) are found during the run of the validLicenseKey task. 

# How to Test


KayLotura2 (approved):
It's kind of surreal to read go, I don't quite know how all the intricacies work, but with a little effort I can mostly make out what's going on. Reminds me of romance languages kind of. 

Anyhow! As far as I can tell, this looks good to me.

pmillspdx (approved):
These are a bunch of huge changes, but it all flows together nicely. Well done!

:shipit:

PR #88, opened 2021-09-20
update ohi case validation for ohis v4

carlossscastro:
Small update to json validation of OHIS as the recommended way is to pass keys in uppercase.
This PR allow json validation to be compatible with the new and old format simply by lowercasing keys when comparing


cade-conklin (approved):
Implementation follows the logic of having the string coming in uppercase.  Looks good to me!

PR #89, opened 2021-09-24
Node v10 is now no longer compatible with New Relic Node.js Agent ver…

cade-conklin:
…sion 8.0.0.0+

# Issue
As stated [here](https://newrelic.atlassian.net/browse/IHUB-309), support for Node v10 is EOL for New Relic  Node.js Agents 8.0.0.0+.  This means that we need to change the compatibility functions of the Diagnostics tool in order to account for those changes.

# Goals
The goal is to change the compatibility for Node that is laid out in Diag.  This means going into our file where all of the version compatibility is stored and making a strict cutoff for the agent supportability for Node v10.  

# Implementation Details
In `tasks/compatibiltyVars/version.go`, we changed the `NodeSupportedVersions` for Node v10 to map to a range of supported New Relic Node.js Agent versions.  Previously, Node v10 mapped to `4.6.0+`, but now it maps to `4.6.0-7.*`.  This means it will support and Node.js Agent version below 8.0.0.0 (where it was EOL'd).  In order to make sure these changes worked, two unit tests were added to `tasks/node/env/versionCompatibility_test.go`.  These made sure that Node v10 with any Node.js Agent in the specified range would return a failure, while Node v10 with any Node.js Agent 8.0.0.0+ would return a success.

# How to Test
In `tasks/node/env/versionCompatibility_test.go` run the package tests to verify they pass.

### DIFF
--- a/tasks/compatibilityVars/versions.go
+++ b/tasks/compatibilityVars/versions.go
@@ -65,7 +65,7 @@ var SupportedForJavaAgent4 = map[string][]string{
 
 var NodeSupportedVersions = map[string][]string{
 	"12": []string{"6.0.0+"},
-	"10": []string{"4.6.0+"},
+	"10": []string{"4.6.0-7.*"},
 }
 
 //https://docs.newrelic.com/docs/agents/net-agent/getting-started/net-agent-compatibility-requirements-net-framework#net-version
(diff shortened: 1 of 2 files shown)


pmillspdx (approved):
Makes sense to me.

:shipit:

PR #100, opened 2022-05-16
Include Flag for File Upload

cade-conklin:
# Issue
We have no way for customers to add files they need included in the `nrdiag-output.zip`

# Goals
We need to add a `-include` to add files/directories up to 4GB to the `nrdiag-output.zip`

# Implementation Details
We added the `-include` flag which runs through the path to first verify it, then make sure it can add files under that path to the directory.  Files of type `.exe` are not included.

# How to Test
`go test ./...`


daffinito (approved):
Looks great to me!

PR #108, opened 2022-05-19
feat: Build for arm64 and optimize binary size

daffinito:
# Issue

We don't support arm64.

# Goals

Support arm64

# Implementation Details

Added build commands to build for arm64 for all 3 os's
Added `-s -w` to ldflags to remove debug info/symbols and reduce binary size (resulted in a 29.5% reduction in size!)

# How to Test

I have an arm based ec2 instance I tested with


cade-conklin (approved):
This is necessary

Issue #26, opened 2020-11-05
Bubble up to the customer that multiple PHP agents can be an issue


uturunku1:
## Summary

improve PHP agent version task to mention which version we found and if there are multiple agents to direct them to this URL:https://discuss.newrelic.com/t/relic-solution-php-agent-not-reporting-web-data/53291
So the customer can understand and fix the issue

The logic for this task lives here: https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/php/agent/version.go

PR #109, opened 2022-05-19
feat: PHP/Agent/Version - update warning if there are multiple versions

daffinito:
# Issue

https://github.com/newrelic/newrelic-diagnostics-cli/issues/26

# Goals

Update PHP version check to include versions in output. Include URL when multiple versions are detected

# Implementation Details

Added to task
Updated test

# How to Test

`go test ./tasks/php/agent -cover`

### DIFF
--- a/tasks/php/agent/version.go
+++ b/tasks/php/agent/version.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


cade-conklin (approved):
Looks good to me!

Issue #111, opened 2022-06-16
.NET Core/5+ version check should use dotnet --info instead of dotnet --version


nr-ahemsath:
Reference: https://github.com/newrelic/newrelic-diagnostics-cli/blob/0191f0a4ab09ea73f2050753e5d5d87a2e4087f7/tasks/dotnetcore/env/versions.go#L76

This task for getting the version(s) of .NET Core/5+ installed on a system will fail if only the .NET runtime is installed, without the SDK being installed.  `dotnet --version` only works if the SDK is installed.  Having only the runtime installed is a common case for containerized (Docker/Kubernetes) deployments.  The more generic command for getting .NET version info is `dotnet --info`.  The output might require some more complicated parsing.  Example output:

```
dotnet --info
.NET SDK (reflecting any global.json):
 Version:   6.0.101
 Commit:    ef49f6213a

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  20.04
 OS Platform: Linux
 RID:         ubuntu.20.04-x64
 Base Path:   /usr/share/dotnet/sdk/6.0.101/

Host (useful for support):
  Version: 6.0.1
  Commit:  3a25a7f1cc

.NET SDKs installed:
  2.1.818 [/usr/share/dotnet/sdk]
  3.1.416 [/usr/share/dotnet/sdk]
  5.0.404 [/usr/share/dotnet/sdk]
  6.0.101 [/usr/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.30 [/usr/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.30 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.1.22 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 5.0.13 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 6.0.1 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.30 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.1.22 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
[...]

PR #113, opened 2022-07-05
fix: Update DotNetCore/Env/Versions

daffinito:
# Issue

https://github.com/newrelic/newrelic-diagnostics-cli/issues/111

# Goals

Update DotNetCore/Env/Versions to use `dotnet --info` for version info. fall back on `dotnet --version` then looking at dirs

# Implementation Details

using a basic enum of sources and loop through until we get versions back

# Testing

There were not tests for this task so I wrote some

`go test ./... -cover`


tasks/dotnetcore/env/versions.go
```diff
@@ -68,22 +79,98 @@ func (t DotNetCoreEnvVersions) Execute(options tasks.Options, upstream map[strin
...
-	errorMessage := "Unable to complete this health check because we ran into some unexpected errors when attempting to collect this application's .NET Core SDK version:\n"
+func (t DotNetCoreEnvVersions) checkVersions(envVars map[string]string) ([]string, string) {
+	source := 0
+	errorMessages := ""
+	versions, errMessage, keepGoing := t.checkVersionWithSource(envVars, source)
+	for keepGoing {
+		source++
+		errorMessages += errMessage
+		versions, errMessage, keepGoing = t.checkVersionWithSource(envVars, source)
+	}
```
cade-conklin, line 90: The only thing I might change is calling this function outside of the function and inside it.  Maybe something like:
keepGoing := true
for keepGoing{
   versions, errMessage, keepGoing := t.checkVersionWithSource(envVars, source)
   errorMessages += errMessage
   source++
}

This isn't necessary to change it just makes more sense when I read the code

cade-conklin (dismissed):
I approve!

tasks/dotnetcore/env/versions.go
daffinito: Good call out! I've refactored it to be more readable (I hope)

cade-conklin (approved):
Looks good to me!

PR #115, opened 2022-07-08
feat: Upload files directly to haberdasher

daffinito:
# Issue

https://newrelic.atlassian.net/browse/IHEDIAG-105

# Goals

Upload from the CLI directly to haberdasher, rather than to S3

# Implementation Details

I moved attach to it's own directory/package and added new functions to upload directly to haberdasher. I kept the legacy functions intact just in case they are needed. To use them, you have to set an env variable `NEWRELIC_LEGACY_ATTACH` to true. Example:

`NEWRELIC_LEGACY_ATTACH=true ./nrdiag -a`

# How to Test

I wrote tests for the new functions I added. `go test ./...`


cade-conklin (approved):
I love dependency injection

PR #124, opened 2022-07-19
include infra agent log gz files (#1)

rogercoll:
# Issue

The infra-agent now supports build-in log rotation, which might create additional log files other than the one specified in the configuration. See: https://github.com/newrelic/infrastructure-agent/blob/master/assets/examples/infrastructure/newrelic-infra-template.yml.example#L159

# Goals

Add into nrdiag all infra-agent log files. 

# Implementation Details

Additionally to the log file specified in the log configuration, rotated filenames will be made with the file base name of the main one + date + gz extension (if compression is enabled)

# How to Test

Implementation relies on tasks.FindFiles function

NOTE: Marked as draft till infra-agent 1.28.0 is released.

### DIFF
--- a/tasks/infra/log/collect.go
+++ b/tasks/infra/log/collect.go
@@ -2,6 +2,8 @@ package log
 
 import (
 	"fmt"
+	"path/filepath"
+	"strings"
 
 	"github.com/newrelic/newrelic-diagnostics-cli/tasks"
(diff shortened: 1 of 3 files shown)


daffinito (approved):
👍 Looks great, thanks again for submitting these PRs!

cade-conklin (approved):
Thank you again!

PR #125, opened 2022-07-20
Fix infra-agent verbose enabled

rogercoll:
# Issue

The infra-agent now supports a new log configuration option rather than the `verbose` one. Verbose will be supported for the following two years, meanwhile some clients can start using the new configuration. See: https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings/#logging-variables

# Goals

Check if verbose (also known as debug or trace level) is enabled using new the new log configuration.

# Implementation Details

Check if new configuration is set first.


cade-conklin (approved):
Great work, thank you!

PR #140, opened 2023-02-27
fix: Updated linux integration test env

daffinito:
# Issue

Integration tests are failing

# Implementation Details

Updated the linux integration tests to use x64 instead of x86, and bumped the ubuntu image up to 22.04. 

# How to Test

`./scripts/integrationTests.sh`


cade-conklin (approved):
Passing locally for me!

PR #142, opened 2023-02-27
Dotnet/Profiler/TLSRegKey

cade-conklin:
# Issue
https://docs.newrelic.com/docs/apm/agents/net-agent/troubleshooting/no-data-appears-after-disabling-tls-10/#windows-registry
# Goals
Ensure that a customer can see data with proper TLS settings
# Implementation Details
Get the `SchUseStrongCrypto` registry settings and check if they are enabled.  If they are not, return a task failure.  Otherwise check if TLS 1.2 `DisabledByDefault` is set to 0 and `Enabled` is set to 1.  If it is, return a task success, otherwise a failure.
# How to Test
- Open a Windows VM
- Make sure you have the .NET agent installed
- Build a usable .exe with `go build` in the root directory
- Run that .exe with the .NET suite `.\newrelic-diagnostics-cli.exe -s dotnet`
- Ensure the output matches the registry settings
- Paths that the registry uses: 
```
var tlsRegKeyPath = `SYSTEM\CurrentControlSet\Control\SecurityProviders\SCHANNEL\Protocols\TLS 1.2\Client`
var SchCryptoRegKeyPath_1 = `SOFTWARE\WOW6432Node\Microsoft\.NETFramework\v4.0.30319`
var SchCryptoRegKeyPath_2 = `SOFTWARE\Microsoft\.NETFramework\v4.0.30319`
```
- TLS has two values `Enabled` and `DisabledByDefault`.  Change those between 1 and 0
- SchCrypto path has one value each: `SchUseStrongCrypto`.  Change those between 1 and 0.
- If both `SchUseStrongCrypto` are set to 1, TLS `Enabled` is set to 1, and TLS `DisabledByDefault` is set to 0, then you should have a success returned


daffinito (approved):
LGTM!

PR #148, opened 2023-03-14
Permalinks to run

cade-conklin:
# Issue
There is no permalink to the Diag CLI Output in New Relic's UI
# Goals
To show where a customer can view the output of the Diag CLI runs
# Implementation Details
Return URL based on run conditions
# How to Test
go test ./...


daffinito (approved):
Smoke tested and works great!

PR #150, opened 2023-03-21
feat: Release platform specific archives to download.newrelic.com

daffinito:
# Issue

We need platform specific archives to support a 1 line installation method

# Goals

To update the upload.sh script to create the needed archives

# Implementation Details

I updated the upload.sh script to create the needed archives. I also added logic to the code for deleting the old releases, it no longer uses the Last Modified date but rather just keeps the current version and previous version (as defined in the releaseVersion.txt file), and deletes anything else.

# How to Test

1. Make sure your local aws cli  is pointing at the [REDACTED]'s AWS account. There is a test bucket there. It currently has releases 2.3.3 and 2.3.4 there.
2. Update the releases in `releaseVersion.txt` to something like

```
majorMinor=2.3
buildNumber=5
currentReleaseVersion=2.3.5
prevReleaseVersion=2.3.4
```

3. Run

```
S3_BUCKET=[REDACTED] BUILD_NUMBER=5 ./scripts/upload.sh
```

4. Check the [REDACTED] bucket and make sure it has 2.3.5 and 2.3.4 releases, nrdiag_latest.zip, and version.txt only:

```
aws s3 ls s3://[REDACTED]/nrdiag/
```

5. Open each archive (they should be in ./nrdiag_scratch_dir/nrdiag/), and ensure the contents look right.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 set -e
 
+TEMPNAME=newrelic-diagnostics-cli
 EXENAME=nrdiag
 
 mkdir -p bin/mac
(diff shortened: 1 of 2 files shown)


cade-conklin (approved):
Looks good to me! Smoke tested and everything!

PR #152, opened 2023-03-28
feat: Install script for linux

daffinito:
# Issue

IHEDIAG-311

# Goals

Create an installation script for nrdiag on Linux to facilitate 1 line install and run

# Implementation Details

The script checks to make sure it's being run on Linux and either x64 or arm64. Linux x86 is no longer supported by nrdiag.

The script also checks for needed prereqs: cut, tar, gzip, sudo, curl

For curl, if it isn't installed, it will ask if the user wants to install it and try to install based on the distro.

By default it installs nrdiag to the pwd, but you can specify a specific directory with `DESTDIR`

The installation goes to a tmp dir then is moved to the final destination. The tmp dir is cleaned up unless there is an error, it's left intact if there's an error to assist with debugging.

The binary is just named `nrdiag` in the end, it removes the `_x64` or `_arm64`.

It only installs the latest version, version is not configurable.

# How to Test

1. Copy the install script to your test host
  * I used scp `scp -i "your-cert" ./scripts/install/install.sh your-host:/tmp`
2. Set the BASE_URL to the test location (ask me for this in slack)
  * `export BASE_URL="https://......com"`
3. Run it
  * `curl -Ls file:///tmp/install.sh | bash && sudo ./nrdiag`
  * feel free to add a task or suite to the command, ie: `curl -Ls file:///tmp/install.sh | bash && sudo ./nrdiag -s node`

### DIFF
--- a/scripts/install/install.sh
+++ b/scripts/install/install.sh
@@ -1,0 +1,143 @@
+#!/usr/bin/env bash
+
+# Install the New Relic Diagnostics CLI.
+# https://github.com/newrelic/newrelic-diagnostics-cli
+#
+# Dependencies: curl, cut, tar, gzip
+#
+# The binary location can be passed in via DESTDIR, otherwise it installs to pwd.
+#
+
+set -e
+
+# ------------------------- INIT
+
+UNAME=($(uname -ms))
+PLAT=${UNAME[0]}
+ARCH=${UNAME[1]}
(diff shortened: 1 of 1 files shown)


cade-conklin (approved):
Looks good to me!

punkael (approved):

PR #160, opened 2023-04-24
chore: Remove urls from testing

daffinito:
# Issue

https://newrelic.atlassian.net/browse/IHEDIAG-335

# Goals

Remove URLs that we don't own from tests

# Implementation Details

Removed the URLs

# How to Test

`go test ./...`


cade-conklin (approved):
LGTM

PR #162, opened 2023-05-02
Python3 and Pip3 Check

cade-conklin:
# Issue
We did not use `python3` or `pip3` to check dependencies
# Goals
Make sure that python tasks didn't fail because no python was installed
# Implementation Details
Changed the `Python/Env/Version` and `Python/Env/Dependencies` to check for `python3 --version` and `pip3 freeze` to ensure that if they don't have python 2 installed, the tasks would still run.  It will return an error if neither version of Python is installed, a warning if only one of the is (and a summary of what was found with the successful command), and a success if both commands run.
# How to Test
Run the python suites and check the output.  Unit tests account for all logic

### DIFF
--- a/domain/entity/pythonVersion.go
+++ b/domain/entity/pythonVersion.go
@@ -1,5 +1,0 @@
-package entity
-
-type PythonVersion struct {
-	Enabled int
-}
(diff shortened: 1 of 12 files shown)


daffinito (approved):
LGTM!

PR #164, opened 2023-05-05
feat: Python Task Update

cade-conklin:
# Issue
Python was checking config files for agent version
# Goals
To check output of pip freeze to see if new relic is installed and compare that against list of python versions that are installed in the user's system
# Implementation Details
Updated `Python/Env/Version`, `Python/Env/Dependencies`, `Python/Requirements/PythonVersion` and their respective unit tests.
# How to Test
Run the python suites and check that the `Python/Requirements/PythonVersion` output does not fail if you have python and the agent installed.  Also `go test ./...`


daffinito (approved):
LGTM!

PR #171, opened 2023-06-08
feat: Add options for EU region

daffinito:
# Issue

https://newrelic.atlassian.net/browse/IHEDIAG-409

# Goals

Add the ability to upload CLI runs to EU accounts

# Implementation Details

Added a command line flag `-r/-region` which accepts either US or EU:

```
-r us
-r eu
```

It is not case sensitive.

Also checks for env variable NEW_RELIC_REGION and uses that if present.

If both the flag and the env variable are present, the flag takes priority.

The default if neither is provided  or if an invalid region is provided is to use US region

# How to Test

I wrote tests for the new functions. 

```
go test ./...
```

Can also be smoke tested by building with BUILD_NUMBER, PROD_ATTACHMENT_ENDPOINT, PROD_USAGE_ENDPOINT, PROD_HABERDASHER_URL, EU_ATTACHMENT_ENDPOINT, EU_HABERDASHER_URL, and EU_USAGE_ENDPOINT all set appropriately.


cade-conklin (approved):
LGTM!

PR #173, opened 2023-07-11
feat: Update workflow and release version

daffinito:
Updates the release workflow to include the EU endpoints to fix the bug in 2.6.0 that caused the `No Haberdasher base URL set. Defaulting to localhost` error

also updating the releaseVersion.txt file at the same time for the new release.


cade-conklin (approved):
👍

PR #175, opened 2023-07-13
fix: Diagnostics CLI Attach only when flags present

cade-conklin:
# Issue
https://newrelic.atlassian.net/jira/software/c/projects/IHEDIAG/boards/2062?modal=detail&selectedIssue=IHEDIAG-457
# Goals
Only prompt the user to attach when they have provided a flag that indicates they want to attach
# Implementation Details
Wrap the prompt in a check for the attachment flags
# How to Test
Smoke test with no attachment flags, `-a`, and `-api-key`

### DIFF
--- a/processTasks.go
+++ b/processTasks.go
@@ -211,12 +211,13 @@ func processUploads() {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


daffinito (approved):
LGTM

PR #177, opened 2023-07-17
fix: Debug instead of info

cade-conklin:
# Issue
Error message in base when running any task
# Goals
Suppress error message
# Implementation Details
Changed info to debugs
# How to Test
Smoke test


daffinito (approved):
LGTM

PR #180, opened 2023-08-16
feat: Diag CLI 3.0.0

daffinito:
# Issue

Adding support for uploads up to 4 GB in size
Cleaning up the output

# Goals

Release version 3.0.0 of the cli

# Implementation Details

Cleaned up some unnecessary white space in the output and added a couple lines to the output when creating the zip and starting the upload. Also added the speed to the upload.
Updated the build script so that mac arm64 is actually built for arm64
Updated version to 3.0.0


cade-conklin (approved):
LGTM!

PR #182, opened 2023-08-25
feat: Run scripts

daffinito:
# Issue

https://newrelic.atlassian.net/browse/IHEDIAG-492

# Goals

Create a framework to run a script and gather the output and have it included in the output json/zip

# Implementation Details

The scripts are located in `scriptcatalog/scripts/`. They have an associated yml file located in `scriptcatalog/`

The CLI uses the github api to download them. 

Use:
- `nrdiag -list-scripts` to list the scripts
- `nrdiag -script example` to view a script
- `nrdiag -script example -run` to run a script with no tasks. This will generate a json/zip file as well for upload
- `nrdiag -script example -run -script-flags "-flag 1 -another"` to run a script with some flags. 
- `nrdiag -script example -run -s node` this will run the script and the node suites.

# How to Test

If you want to smoke test, you will have to add a param to the MakeRequest function in catalog.go. Add this right before the return statement on line 149:

```
wrapper.Params.Add("ref", "daffi/scripts")
```

Then you can `go build` and use the above commands to test

Also: `go test ./...`


cade-conklin (approved):
LGTM!

Issue #184, opened 2023-08-29
JSON and filelist aren't added to zip if using -output-path


daffinito:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

When redirecting output with `-output-path`, the json file and the filelist file are not added to the zip file

## Steps to Reproduce

run the cli with `-output-path` and review the zip

## Expected Behavior

both files should be added to the zip

PR #187, opened 2023-08-30
fix: Issue 184

daffinito:
# Issue

https://github.com/newrelic/newrelic-diagnostics-cli/issues/184

# Goals

Ensure filelist and json are in zip when using -output-path

# Implementation Details

Ensured to use filepath.Join to build paths

# How to Test

build it and run a task with -output-path and check the resulting zip to ensure it contains the filelist and json


cade-conklin (approved):
👍

PR #188, opened 2023-08-31
fix: Log gathering logic

daffinito:
# Issue

https://newrelic.atlassian.net/browse/IHEDIAG-480
Logs are not being collected as expected

# Goals

Ensure all logs expected are collected

# Implementation Details

This issue had 2 parts:

- Base/Log/Collect added files to `FilesToCopy` that didn't exist
- The loop logic in output.copyFilesToZip which adds files to the zip would `return` on any error, instead of `continue`. This would prevent any of the log files in the slice after the one that caused the error from being added to the zip.

To solve the problem I:

- Updated Base/Log/Collect to ensure files exist before adding them to `FilesToCopy`
- Updated  output.ProcessFilesChannel to also ensure files exist before adding them to the zip/filelist
- Updated the loop logic in output.copyFilesToZip to continue processing if it encounters an error

# How to Test

smoke test it and ensure all the log files you expect are collected

---

I also ran `gofmt -s -w .` on the whole project, so there are a lot of little formatting changes

### DIFF
--- a/core.go
+++ b/core.go
@@ -146,7 +146,6 @@ func main() {
 		// creates the output file
 		output.WriteOutputFile(outputResults, scriptData)
 
-
 		// copy our output file(s) to the zip file
 		output.CopyOutputToZip(zipfile)
 		if scriptData != nil {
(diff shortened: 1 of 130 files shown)


cade-conklin (approved):
🥇

PR #190, opened 2023-09-08
fix: Check for actual usage endpoint

cade-conklin:
# Issue
Usage endpoint defaults to localhost
# Goals
Use correct usage endpoint
# Implementation Details
Check directly for usage endpoint
# How to Test
Run

### DIFF
--- a/usage/usage.go
+++ b/usage/usage.go
@@ -107,7 +107,6 @@ func genInsertKey(runID string) string {
 
 // SendUsageData generates and POSTS NRdiag usage data to the Haberdasher service
 func SendUsageData(results []registration.TaskResult, runID string) {
-
 	preparedResults := prepareResults(results)
 	preparedConfig := config.Flags.UsagePayload()
 	preparedMeta := prepareMeta(results, runID)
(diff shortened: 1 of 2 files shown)


daffinito:
As I was reviewing this, I realized that the root issue is an order of operations issue. Previously (before the [region check PR](https://github.com/newrelic/newrelic-diagnostics-cli/pull/171)), `config.UsageEndpoint` was being set by the build script, so it is set before the `main` function even runs. This meant that when the global variable `usage.usageEndpoint` was being set to `config.UsageEndpoint`, `config.UsageEndpoint` was already set to the correct endpoint.

After the region PR, `config.UsageEndpoint` is now being set in `config.ParseFlags()`, so it is being set AFTER the global variable `usage.usageEndpoint` has already been set, causing `usage.usageEndpoint` to be set to nothing. 

The solution is to just not use a global variable for `usage.usageEndpoint` as there's no reason to use one for it. I moved the variable creation into the `usage.SendUsageData` and it's all being set in the correct order again.

daffinito (approved):

PR #195, opened 2023-09-27
feat: Infra Debug log improvements

daffinito:
# Issue

* Infra logs are not always being collected when using the infra:debug suite
* Infra debug log collection needs to be able to be gracefully interrupted

# Goals

* Try to make sure Infra logs are gathered whenever possible
* Add a way to gracefully interrupted the 5 minute wait for logs with infra:debug

# Implementation Details

1: `Base/Log/Copy` had a bug where it would try to copy a file with a blank name:
```
{
  "FileName": "",
  "FilePath": "",
  "Source": {
	  "FoundBy": "Found by looking at standard locations",
	  "KeyVals": null,
	  "FullPath": ""
  },
  "IsSecureLocation": false,
  "CanCollect": false,
  "ReasonToNotCollect": "stat : no such file or directory"
}
```

I resolved this in `tasks/base/log/logHelpers.go` 

2: `Infra/Agent/Debug` refactored to depend on `Infra/Log/Collect` instead of `Base/Log/Copy`. This should be more reliable and gives us more control over the result summaries and help us narrow down any further issues with this task.

Also added the ability to hit Ctrl-C to gracefully cancel the 5 minute timer and keep moving with the rest of the run.

3: `Infra/Log/Collect` updated to include env variables and default locations.

4: `Base/Env/RootUser` is a new task for Linux/Darwin that checks if the user running the CLI is root/using sudo. Added this as I suspect some of the issues with gathering infra logs may be permissions based.

5: The jvm task wasn't altered, I just split the registration up for linux and mac to separate files for more granular control over where tasks get registered. (I wanted the SELinux task to only register on linux but both to have the RootUser task)

# How to Test

`go test ./...`


cade-conklin (approved):
:shipit:

Issue #192, opened 2023-09-15
connection-test.newrelic.com URL Is not active 


pzimno:
```
curl https://connection-test.newrelic.com
curl: (7) Failed to connect to connection-test.newrelic.com port 443 after 9 ms: Couldn't connect to server
```
## Description

The connection test fails during a diag check but the URL seems to be down when attempting to hit it from multiple locations and trying to open a TCP connection to port 443 on the server.

## Steps to Reproduce

Run the diag tool and check connection test output or curl https://connection-test.newrelic.com

## Expected Behavior

https://connection-test.newrelic.com Succeeds with a 200 OK

## Relevant Logs / Console output

```
Failure - Base/Collector/ConnectTLS
There was an error connecting to connection-test.newrelic.com.
Please check network and proxy settings and try again or see -help for more options.
Error = Get "https://connection-test.newrelic.com/": dial tcp 162.247.242.43:443: connectex: No connection could be made because the target machine actively refused it.
See https://docs.newrelic.com/docs/new-relic-solutions/get-started/networks for more information.
```

## Your Environment

Tested with a Windows and Linux based environment

daffinito:
hi @pzimno , thank you for your patience! It looks like `connection-test.newrelic.com` is no longer going to be available, so we have removed the task `Base/Collector/ConnectTLS`. There are other tasks that check the connection, `Base/Collector/ConnectUS` or `Base/Collector/ConnectEU` depending on your region. The endpoint used by those tasks (`https://collector[.eu].newrelic.com/status/mongrel`) is only accessible if TLS is 1.2 or higher, so if those succeed, it can be safely assumed that TLS is 1.2 or higher.

Thanks again for reporting this to us!

PR #197, opened 2023-10-02
fix: Remove ConnectTls, update ConnectUs/Eu

daffinito:
# Issue

https://github.com/newrelic/newrelic-diagnostics-cli/issues/192

# Implementation Details

We have removed `Base/Collector/ConnectTLS` due to https://connection-test.newrelic.com being deprecated

We updated `Base/Collector/ConnectUS` and `Base/Collector/ConnectEU` to use the mongrel endpoints again:

* https://collector.newrelic.com/status/mongrel
* https://collector.eu.newrelic.com/status/mongrel

These endpoints are not available unless the connection is using TLS 1.2+, so if they succeed, TLS shouldn't be an issue.

# How to Test

`go test ./...`


cade-conklin (approved):
LGTM!

PR #207, opened 2023-11-16
fix: No longer collect executable files

daffinito:
# Issue

IHEDIAG-649
Infra log collection can collect .exe files

# Goals

Ensure no executable files are being collected

# Implementation Details

I updated the regex that is used in the Infra/Log/Collect task to be more specific. If the infra config has a log path that ends with `newrelic-infra.log`, the regex would be
* OLD:  `^newrelic-infra*`
* NEW: `^newrelic-infra.*(.gz|.zip|.log)`

I updated the FileCopyEnvelope struct with a function to check if a file is executable. For windows it just checks if the suffix is `.exe`. For Linux/Mac, it checks the file permissions as described in [this stack overflow](https://stackoverflow.com/questions/60128401/how-to-check-if-a-file-is-executable-in-go).

I updated the paths we search for config files to include the paths for Infra on Mac installed via brew. Taken from [this doc](https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/macos-installation/install-infrastructure-monitoring-agent-macos/#homebrew-install).

# How to Test

go test ./...

I smoke tested by:
* Running the Infra suite and confirming it collected the newrelic-infra.log
* Moving the newrelic-infra.log and replacing it with an executable file
* Running the infra suite again, and confirming that `Skipping executable file:` is in the debug logging and that the log file is not in the zip

### DIFF
--- a/output/output.go
+++ b/output/output.go
@@ -145,6 +145,7 @@ func ProcessFilesChannel(zipfile *zip.Writer, wg *sync.WaitGroup) {
 		log.Debug("Copying files from result: ", result.Task.Identifier().String())
 
 		for _, envelope := range result.Result.FilesToCopy {
+			log.Debug("Copying file: ", envelope.Path)
 			if !tasks.FileExists(envelope.Path) {
 				log.Debugf("File does not exist, skipping: '%s'\n", envelope.Path)
 				continue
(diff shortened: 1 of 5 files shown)


cade-conklin (approved):
:shipit:

PR #212, opened 2024-01-11
feat: High Security Mode Refactor

cade-conklin:
# Issue
HSM Refactor
# Goals
Refactor High Security Mode task to only check env vars and configuration files.  No comparison no longer happens in the CLI and instead can be sent up to their account to compare against account settings in the UI.
# Implementation Details
- Took out license key dependency from HSM task.
- Only checks env vars and configuration files for high security mode
- (Bugfix) Check config file settings for HSM
- Output info if settings are found
# How to Test
go test ./..
smoke test

### DIFF
--- a/output/output.go
+++ b/output/output.go
@@ -272,10 +272,16 @@ func WriteLineResults() []registration.TaskResult {
 	var filtered [6]int
 
 	var outputResults []registration.TaskResult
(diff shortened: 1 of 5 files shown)


daffinito (approved):
This is the way!

PR #219, opened 2024-03-22
fix(NR-231490): Update Infra/Config/IntegrationsMatch: make integration_name optional

daffinito:
# Issue

NR-231490

# Goals

Update Infra/Config/IntregrationsMatch to make integration_name optional if the configuration is v4.

# Implementation Details

Added logic to not check for integration_name if the file is v4.

I also noticed this failure:

```
Failure - Infra/Config/ValidateJMX
Unexpected results for jmx-config.yml: invalid configuration found: collection_files not set
```

The field collection_files is now in caps. Here is what the default yml looks like:

```
integrations:
- name: nri-jmx
  env:
    COLLECTION_FILES: "/etc/newrelic-infra/integrations.d/jvm-metrics.yml,/etc/newrelic-infra/integrations.d/tomcat-metrics.yml"
    JMX_HOST: jmx-host.localnet
    JMX_PASS: admin
    JMX_PORT: "9999"
    JMX_USER: admin

    # New users should leave this property as `true`, to identify the
    # monitored entities as `remote`. Setting this property to `false` (the
    # default value) is deprecated and will be removed soon, disallowing
    # entities that are identified as `local`.
    # Please check the documentation to get more information about local
    # versus remote entities:
    # https://github.com/newrelic/infra-integrations-sdk/blob/master/docs/entity-definition.md
    REMOTE_MONITORING: "true"
  interval: 15s
  labels:
    env: staging
```

So I updated the `FindKeys()` config helper function to be case insensitve.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ hog.out
 .go-version
 #coverage file
 coverage.out
+# local history 
+.history
--- a/tasks/infra/config/integrationsMatch.go
+++ b/tasks/infra/config/integrationsMatch.go
@@ -74,8 +74,6 @@ func (p InfraConfigIntegrationsMatch) Dependencies() []string {
 
 // Execute - The core work within each task
 func (p InfraConfigIntegrationsMatch) Execute(options tasks.Options, upstream map[string]tasks.Result) tasks.Result {
-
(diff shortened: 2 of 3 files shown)


cade-conklin (approved):
LGTM!

Issue #226, opened 2024-06-05
Agent Reports Running but Not Receiving Data within New Relic UI   


nls91TAG:
**Description:**

After installing the New Relic Infrastructure agent and waiting for it to initialize on the New Relic API, it appears that the infrastructure data is not being passed in. I checked the status of the agent using Get-Service newrelic-infra).Status, which returns as running. I used the Diagnostics CLI for further debugging:

**Error - Infra/Config/ValidateJMX**
Diagnostics CLI was unable to complete this health check because we ran into an unexpected type assertion error.

**Error - Infra/Env/NrjmxMbeans**
Diagnostics CLI was unable to complete this health check because we ran into an unexpected type assertion error.

**Warning - Base/Config/AppName**
No New Relic app names were found. Please ensure an app name is set in your New Relic agent configuration file or as a New Relic environment variable (NEW_RELIC_APP_NAME). Ignore this warning if you are troubleshooting for a non APM Agent.

**Error - Java/JVM/VendorsVersions**
Diagnostics CLI was unable to detect any running Java processes be
cause we ran into an error: could not get Name: A device attached to the system is not functioning.

**Steps to Reproduce:**

After installing the .MSI installer agent on the server and waiting 5 minutes, I am unable to observe any data when checking the host summary.

**Expected Behavior**

Infrastructure data to transferred into the New Relic API.

**Relevant Logs / Console output**

*link removed by New Relic*

daffinito:
Hi @nls91TAG, thanks for reaching out! 

I went ahead and removed the link to your `nrdiag-output.zip` - it contained your license key.

For your Infrastructure Agent issues, I recommend running nrdiag again with the `infra:debug` suite (`nrdiag -s infra:debug`), then providing that to support for assistance. You can find support resources here:

* https://one.newrelic.com/redirect/support
* https://forum.newrelic.com

For these errors:

```
Diagnostics CLI was unable to complete this health check because we ran into an unexpected type assertion error.
```

I will look into these and determine why we're getting unexpected types in these tasks.

daffinito:
hi @nls91TAG, it looks like the unexpected type assertion errors were due to running the 32 bit version of the CLI. Unless you are specifically troubleshooting a 32 bit APM application, we recommend running the 64 bit version of the binary.

Regarding this other error from the Java/JVM/VendorsVersions task:

```
Diagnostics CLI was unable to detect any running Java processes be
[...]

PR #228, opened 2024-06-20
fix(226): Handles errors in FindFiles and FindProcessByName and upgades gopsutil dep

daffinito:
#  Issue

https://github.com/newrelic/newrelic-diagnostics-cli/issues/226

# Goals

Stop Java/JVM/VendorsVersions from reporting this error: Diagnostics CLI was unable to detect any running Java processes because we ran into an error: could not get Name: A device attached to the system is not functioning.

Update dependencies

Fix regex related panic 

# Implementation Details

Instead of returning an error when we can't determine a process's name, I changed to so that we log it to debug logging but keep processing the rest of the pid list.

Changed the regex.MustCompile to regex.Compile and logged the error.

# How to Test

Unit tests


cade-conklin (approved):
LGTM!

PR #231, opened 2024-07-01
fix(NR-268332): Changed infra/agent/version to warning if cant validate publish date

daffinito:
# Issue

The Infra/Agent/Version task does a check to ensure the agent version is not out of date by pulling the data from the github api. If that connection fails, the task returns an error. This causes Infra/Agent/Debug to not gather logs and to report that the agent is not installed when it actually is.

# Goals

Have Infra/Agent/Debug run even if the agent version can't be checked against github.

# Implementation Details

Changed Infra/Agent/Version to return a warning when it can't validate the agent version isn't outdated.
Changed Infra/Agent/Debug to keep going even if Infra/Agent/Version had a warning.

### DIFF
--- a/tasks/infra/agent/debug.go
+++ b/tasks/infra/agent/debug.go
@@ -76,7 +76,7 @@ func (p InfraAgentDebug) Execute(options tasks.Options, upstream map[string]task
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


linbaker (approved):
Thanks for fixing!

PR #237, opened 2025-04-02
feat: replace k8s Infra agent gathering with the new agent control resources/logs/helm-flux

alvarocabanas:
Create a new task for k8s/agent-control gathering (and replace k8s Infra task):

- All Configmaps in given namespace
- All Pods
- All Deployments
- All Daemonsets
- helmreleases.helm.toolkit.fluxcd.io
- helmcharts.source.toolkit.fluxcd.io
- helmrepositories.source.toolkit.fluxcd.io
- helm list -a
- All logs from agent-control and flux

You can call:

`nrdiag -y` --> All
`nrdiag -suites K8s` --> Only k8s
`nrdiag -suites K8s-agent-control` --> All K8s agent-control


tasks/k8s/helm/flux_repositories.go
```diff
@@ -0,0 +1,72 @@
...
+	var (
+		res []byte
+		err error
+	)
+
+	namespace := options.Options["k8sNamespace"]
+	res, err = p.runCommand(namespace)
+	if err != nil {
+		return tasks.Result{
+			Summary: "Error retrieving pods: " + err.Error(),
```
rubenruizdegauna, line 38: Shouldn't this need to be `Error retrieving Flux Helm repositories:`?

config/config.go
paologallinaharbur, line 1: is there somewhere an "help message" explaining to the user what is going to collect?

processOptions.go
```diff
@@ -248,6 +248,11 @@ func processOverrides() (tasks.Options, []override) {
 		options.Options["YesToAll"] = "true"
 	}
 
+	if config.Flags.K8sNamespace != "" {
```
paologallinaharbur, line 251: if not set the default it is an empty string right?

```
	defaultString := ""
	flag.StringVar(&Flags.K8sNamespace, "k8s-namespace", defaultString, "Specify a namespace to use when executing the kubectl command")

```

tasks/k8s/helm/flux_releases.go
```diff
@@ -0,0 +1,72 @@
...
+			Summary: "Error retrieving Flux Helm Releases: " + err.Error(),
+			Status:  tasks.Error,
+		}
+	}
+
+	stream := make(chan string)
+	go tasks.StreamBlob(string(res), stream)
+
+	return tasks.Result{
+		Summary:     "Successfully collected Flux Helm Releases",
```
paologallinaharbur, line 47: Whould we place all flux related stuff under a Flux package instead of an Helm one?

registration/registerTasks.go
```diff
@@ -121,8 +123,10 @@ func init() {
 	baseAgent.RegisterWith(Register)
 	nodeRequirements.RegisterWith(Register)
 	rubyRequirements.RegisterWith(Register)
-	k8sInfra.RegisterWith(Register)
 	k8sEnv.RegisterWith(Register)
+	k8sResources.RegisterWith(Register)
+	k8slogs.RegisterWith(Register)
+	K8sHelm.RegisterWith(Register)
```
paologallinaharbur, line 129: now that all of them are registered how are you calling them?

paologallinaharbur (commented):
The only problem I see is with the namespace and that everything is AC related, but the package is not.

Moreover where in the code are we deciding which resources to are we collecting?

config/config.go
```diff
@@ -260,14 +260,6 @@ func ParseFlags() {
...
 	}
 
-	if Flags.K8sNamespace != "" {
-		Flags.Override = "K8s/Infra/Config.namespace=" + Flags.K8sNamespace + "," + Flags.Override
-		Flags.Override = "K8s/Infra/Daemonset.namespace=" + Flags.K8sNamespace + "," + Flags.Override
-		Flags.Override = "K8s/Infra/Deploy.namespace=" + Flags.K8sNamespace + "," + Flags.Override
-		Flags.Override = "K8s/Infra/Logs.namespace=" + Flags.K8sNamespace + "," + Flags.Override
-		Flags.Override = "K8s/Infra/Pods.namespace=" + Flags.K8sNamespace + "," + Flags.Override
-	}
-
```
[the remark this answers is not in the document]
alvarocabanas: This was injecting the namespace one by one only to this tasks, now I inject the namespace to all the tasks. That's needed becauser there are some tasks like the ones for the helm charts that form the name dinamically, but also because there are so many more tasks needing it so it doesn't make sense to put a bit list here because passing the namespace to all is not a problem.

config/config.go
alvarocabanas: If you execute:
`ndiag  -h tasks `

you can see a list of the tasks that will be executed with their descriptions, including the new k8s ones.


image

tasks/k8s/helm/flux_repositories.go
alvarocabanas: Fixed in last commit

processOptions.go
alvarocabanas: Yes it is.

alvarocabanas: So it gathers everyting from your actual context, maybe it's an unwanted behavior.

config/config.go
alvarocabanas: With the new format for Flux:
image

tasks/k8s/helm/flux_releases.go
alvarocabanas: Done in last commit.

tasks/k8s/helm/flux_repositories.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return tasks.Result{
+		Summary:     "Successfully collected Flux Helm Repositories",
+		Status:      tasks.Info,
+		FilesToCopy: []tasks.FileCopyEnvelope{{Path: "FluxHelmRepositories.txt", Stream: stream}},
+	}
+}
+
+func (p FluxRepositories) runCommand(namespace string) ([]byte, error) {
+	if namespace == "" {
```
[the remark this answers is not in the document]
alvarocabanas: Is that what we want? or we want to gather your current kubectl context. That's what is doing now.

tasks/k8s/logs/logs.go
```diff
@@ -0,0 +1,28 @@
...
+		labelSelector: "app=helm-controller",
+	}, true)
+	registrationFunc(K8sPodLogs{
+		cmdExec:       tasks.CmdExecutor,
+		appName:       "source-controller",
+		labelSelector: "app=source-controller",
+	}, true)
+	registrationFunc(K8sPodLogs{
+		cmdExec:       tasks.CmdExecutor,
+		appName:       "agent-control",
```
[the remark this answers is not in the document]
alvarocabanas: How would you organize it?

config/config.go
paologallinaharbur: ok thanks!

processOptions.go
paologallinaharbur: 🤔 maybe you are right, this is the behaviour that I would expect

tasks/k8s/helm/flux_repositories.go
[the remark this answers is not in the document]
paologallinaharbur: answered above, I think you are right, let's keep it in this way!

tasks/k8s/logs/logs.go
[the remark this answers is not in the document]
paologallinaharbur: I was thinking to have a group AC instead of Flux having both logs and Flux stuff (and everything else related to AC)

Maybe it is not that simple 😕

registration/registerTasks.go
alvarocabanas: I've added this 2 new suites:

image

alvarocabanas: If you execute :
nrdiag -h suites

You can see the different suites.

So you can call:
- ` nrdiag -y` --> All
- ` nrdiag -suites K8s `--> Only k8s
-  `nrdiag -suites K8s-agent-control` --> All K8s agent-control

paologallinaharbur (approved):

daffinito (approved):
Thank you for this contribution!

PR #243, opened 2025-07-02
feat: allow split-namespace gathering for AgentControl

alvarocabanas:
On newest versions of AC 2 namespaces are used the main one and one for the agents.
This task allows adding a flag specifying an the extra namespace where the agent control agents reside.
The Resources will be scrapped from both namespaces,

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -58,6 +58,7 @@ type userFlags struct {
 	Script             string
 	ScriptFlags        string
 	K8sNamespace       string
+	ACAgentsNamespace  string
 	InNewRelicCLI      bool
 }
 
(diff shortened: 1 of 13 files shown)


alvarocabanas:
> LGTM!
> 
> Has this been tested? I ask because I'm not sure what would be the output of the `getResources` function and if it makes sense at the "other end" of the streaming.
> 
> For example, would it be the same if inside `getResources` the `result` was `[][]byte` and inside `Execute` we call `tasks.StreamBlob(string(r)), stream)` for each `[]byte` element in the slice?

Yes, I tested it and works as expected, I thought about parallelizing it more, but it's done sequntially so the resources don't get mixed.

daffinito (approved):

PR #252, opened 2025-08-11
fix(glibc): Ensuring nrdiag uses pure Go

daffinito:
# Issue

When running the build script locally and testing, I do not receive any errors on any linux distros and the binary is statically linked

```
$ ldd nrdiag_x64
	not a dynamic executable
```

However the binaries that are being built by our release pipeline result in the following errors on Rocky Linux 8:

```
./nrdiag: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./nrdiag)
./nrdiag: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./nrdiag)
```

which is caused by them being dynamically linked:

```
$ ldd nrdiag
./nrdiag: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./nrdiag)
./nrdiag: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./nrdiag)
	linux-vdso.so.1 (0x00007ffd281ec000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f6869ace000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f6869ea4000)
```

# Implementation Details

* Moved the CGO_ENABLED=0 setting to the go build calls directly
* Added netgo tag to ensure pure go is used for net library
* Added check to lint.yml to fail PR if any package that uses cgo is used

# How to Test

I was unable to reproduce the issue when building locally, even using tools like [act](https://github.com/nektos/act) to try and reproduce the github actions.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,7 +1,7 @@
 name: golangci-lint
 on:
   pull_request:
-   paths-ignore:
+    paths-ignore:
       - 'README.md'
       - '*_test.go'
 jobs:
@@ -9,7 +9,6 @@ jobs:
     strategy:
       matrix:
         go-version: [1.22.x]
-        # os: [ubuntu-latest]
         os: [ubuntu-latest, macos-latest, windows-latest]
     name: lint
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


nr-linbaker (approved):
LGTM, TY!

PR #263, opened 2026-05-14
feat: Add support for Japan Region

daffinito:
# Goals

Add support for upcoming Japan Region

# Implementation Details

Adds `--region/-r jp` flag to mimic EU region


nr-linbaker (approved):
Looks good to me!

PR #266, opened 2026-07-13
fix(tests): Apply more fixes for Windows Integration test flakiness

daffinito:
## Summary
The `InfrastructureVersionPresentWindows` integration test has been failing on and off. It exercises `Infra/Agent/Version`, which reports `Info` only after calling the GitHub API to check how old the installed agent is. That call is unauthenticated and rate-limited to 60 requests/hour per IP, and the Windows runners share IPs, so the call frequently gets throttled. When it does, the task returns `Warning` instead of `Info` and the test fails. Re-running the same commit passes, which is what pointed to the rate limit rather than a real regression.

## Fix
Instead of depending on that network call, the test now hands the release date directly to the task via the `-o` override the CLI already supports. Nothing changes in production: when the override isn't set, the task calls GitHub exactly as before. The date the test passes in is generated from the current date at image-build time, so it always looks recent and won't silently rot into a failure a year out the way a hardcoded date would.

The version-age logic is already unit-tested, so this doesn't drop meaningful coverage. The integration test goes back to verifying what it's actually for: that the installed agent binary is found and its version is read end to end.


mscotto-nr (approved):
Looks good.

PR #270, opened 2026-07-31
feat: add -agent-control flag to default k8s/ac-agents namespaces

DavSanchez:
## Summary
- Adds a new `-agent-control` flag that pre-fills `-k8s-namespace` (`newrelic-agent-control`) and `-ac-agents-namespace` (`newrelic`) with the defaults for a standard Agent-control Kubernetes installation.
- Either namespace flag, if set explicitly, still takes precedence over the preset — overriding both makes `-agent-control` redundant, which is an acceptable tradeoff for the convenience it provides.
- Only affects the `k8s-agent-control` suite/tasks; has no effect outside Kubernetes (the on-host `AgentControl/*` tasks don't consume these options).

## Test plan
- [x] `go build ./...`
- [x] `go test ./config/...` — existing `Test_userFlags_UsagePayload` updated for the new field, plus new `Test_applyAgentControlDefaults` covering flag-off, both-empty, and all override combinations


vjripoll (approved):
LGTM

daffinito:
There's already 39 flags on the CLI, so I want to be careful about adding more. What if you used `SelectedSuites` here instead, [the way `Base/Agent/EOL` does](https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/base/agent/EOL.go#L67)? `-s k8s-agent-control` already tells us everything `-agent-control` would.

Or is it possible to detect the namespaces? I see [the label in `agentcontrol.go`](https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/k8s/agentcontrol/agentcontrol.go#L36), and [`retrievePodName` in `status_server.go`](https://github.com/newrelic/newrelic-diagnostics-cli/blob/main/tasks/k8s/agentcontrol/status_server.go#L87) already looks pods up by label, so I'm wondering if something like that could work here.

Either way, one thing worth flagging: `k8sNamespace` and `ACAgentsNamespace` are global task options rather than suite-scoped, so this wouldn't only affect the `k8s-agent-control` suite, these defaults would land on `-s k8s` and a bare `nrdiag` too.

